PR #4, opened 2020-11-12
Implement Boruvka's Algorithm

Victor-C-Zhang:
Implementation and small testing loads for Boruvka
Disabling optimizations for local debugging use cases
Changing some types in L0 to accommodate larger inputs


CMakeLists.txt
```diff
@@ -140,18 +140,55 @@ if(NOT Boost_FOUND)
...
   l0_sampling/sketch.cpp
   l0_sampling/update.cpp
+  util.cpp
 )
 target_link_libraries(l0sampling PRIVATE xxHash::xxhash)
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  target_compile_options(l0sampling PRIVATE -O3)
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+  target_compile_options(l0sampling PRIVATE /O2)
+endif()
```
Abi1024, line 169: I'm not sure why we need different compiler options based on the compiler here?

Victor-C-Zhang: Different compilers have different syntax for the optimization flags. This bit of code is copied from the previous iteration, iirc

include/graph.h
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Main algorithm utilizing Boruvka and L_0 sampling.
+   * @return a vector of the connected components in the graph.
+   */
+  vector<set<Node>> connected_components();
+};
+
+class UpdateLockedException : public exception {
+  virtual const char* what() const throw() {
+    return "The graph cannot be updated: Connected components algorithm has "
```
Abi1024, line 38: I don't have much experience with C++'s exception handling. Could you explain this part in more depth?

Victor-C-Zhang: This defines an exception that is thrown if we try to modify the graph after the connected components algo is run. It prints an error string as defined in the return.

include/sketch.h
```diff
@@ -12,14 +12,14 @@ using namespace std;
  */
 class Sketch {
   const long seed;
-  const int n;
+  const unsigned long long int n;
```
tseip: unsigned long long is fine (and correct) but consider using uint64_t in the future. I think it shows intent more clearly. other integer types in C++ aren't as well defined as long longs (which are guaranteed to be at least 64 bits, but not necessarily exactly 64 bits)

include/util.h
```diff
@@ -1,10 +1,34 @@
 #pragma once
 
+typedef long long int ll;
```
tseip: I'm not so sure we want these in the header file unless you want these typedefs accessible everywhere. Maybe an anonymous namespace?

util.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+const ull ULLMAX = std::numeric_limits<ull>::max();
+
+unsigned long long int double_to_ull(double d, double epsilon) {
+  return (unsigned long long) (d + epsilon);
+}
+
+ull nondirectional_non_self_edge_pairing_fn(ull i, ull j) {
+  // swap i,j if necessary
+  if (i > j) {
```
tseip, line 17: use std::swap - it should be as fast or faster than this

util.cpp
```diff
@@ -0,0 +1,55 @@
...
+    i^=j;
+    j^=i;
+    i^=j;
+  }
+  ull jm = j-1ull;
```
tseip, line 20: maybe there's something subtle here that I'm not seeing but isn't all of this equivalent to

const ull max_in = std::max(i, j);
const ull min_in = std::min(i, j);
return min_in * max_in * (max_in / 2ull)

If it is, I think the above will be much faster given that there's no branching.

util.cpp
```diff
@@ -0,0 +1,55 @@
...
+  else jm>>=1ull;
+  if (ULLMAX/j < jm)
+    throw std::overflow_error("Computation would overflow unsigned long long max");
+  j*=jm;
+  if (ULLMAX-j < i)
+    throw std::overflow_error("Computation would overflow unsigned long long max");
+  return i+j;
+}
+
+std::pair<ull, ull> inv_nondir_non_self_edge_pairing_fn(ull idx) {
```
tseip, line 31: It's unclear to me what this function is doing exactly - think about this for a bit and see if you can write it in a clearer way

util.cpp
Victor-C-Zhang: I don’t see how it’s faster: std::swap uses copy-construction, according to `cppreference`
```
template <class T> void swap ( T& a, T& b ) {   T c(a); a=b; b=c; }
```

Is this an empirical result you’ve seen in practice?

util.cpp
```diff
@@ -0,0 +1,55 @@
...
+  else i = idx-j*((j-1ull)>>1ull);
+  return {i, j};
+}
+
+ull cantor_pairing_fn(ull i, ull j) {
+  if (ULLMAX - i < j)
+    throw std::overflow_error("Computation would overflow unsigned long long max");
+  if (j < ULLMAX && ULLMAX - i < j+1ull)
+    throw std::overflow_error("Computation would overflow unsigned long long max");
+  ull am = i+j, bm = i+j+1ull;
```
tseip, line 46: see comment in nondirectional_edge_self_pairing, I think this is more or less the same math

include/util.h
```diff
@@ -1,10 +1,34 @@
...
+
 /**
  * Cast a double to unsigned long long with epsilon adjustment.
  * @param d         the double to cast.
  * @param epsilon   optional parameter representing the epsilon to use.
  */
-unsigned long long int double_to_ull(double d, double epsilon = 0.00000001) {
-  return (unsigned long long) (d + epsilon);
-}
+unsigned long long int double_to_ull(double d, double epsilon = 0.00000001);
```
tseip, line 8: I still don't quite understand what this is meant to do (or rather, why std::round is insufficient)

include/graph.h
```diff
@@ -0,0 +1,44 @@
...
+enum UpdateType {
+  INSERT,
+  DELETE,
+};
+
+typedef pair<Edge, UpdateType> GraphUpdate;
+
+class Graph{
+  const unsigned long long int num_nodes;
+  bool UPDATE_LOCKED = false;
```
tseip: nit: I don't think this has to be all-caps. Nothing else is and I think it's at best misleading (it isn't const)

graph.cpp
```diff
@@ -0,0 +1,83 @@
...
+    delete supernodes[i];
+  delete supernodes;
+  delete representatives;
+}
+
+void Graph::update(GraphUpdate upd) {
+  if (UPDATE_LOCKED) throw UpdateLockedException();
+  Edge &edge = upd.first;
+  // ensure lhs < rhs
+  if (edge.first > edge.second) {
```
tseip, line 28: std::swap

tseip (changes requested):
Had a few comments - see them inline. You don't necessarily need to 'fix' every one (you may disagree with them), but some response would be good

util.cpp
tseip: Even if cppreference mentions that as a reference implementation (I can't find it), almost no library would ever implement it that way, and even if they did, modern compilers would be able to optimize the above significantly better than the xor trick. Also, that doesn't necessarily invoke copy construction - it's likely move construction at best.

See https://godbolt.org/z/54esx8 for a direct comparison in an arbitrary compiler - you can see std::swap generates fewer instructions with any optimization enabled.

tseip: Perhaps even more notably, with NO optimization the reference implementation you give above is actually better:
https://godbolt.org/z/3o5Mnd
And with optimization it reduces to std;:swap

include/sketch.h
Victor-C-Zhang: good point, I'll change that with the other changes in the PR

include/util.h
Victor-C-Zhang: probably true, I'll add them to `cpp` files where I use them

util.cpp
Victor-C-Zhang: interesting, I didn't know that. I'll change that

util.cpp
Victor-C-Zhang: We're potentially working with numbers where these calculations could silently overflow. We want to test if any of these operations would overflow, and report it.

include/util.h
Victor-C-Zhang: Admittedly, this is a hack to get around another, hackier, hack. We want the same effect as integer flooring, but there's a weird quirk in floating point to unsigned types where sometimes a value very close to a whole number `n` will round down to `n-1`. We could compromise by always rounding up, but since this rounding shows up in creating `c*log n` buckets, this may end up increasing our memory footprint by a non-insignificant amount.

To me, this is the least-bad option, though I would like to hear your thoughts on it.

util.cpp
Victor-C-Zhang: this inverts the pairing function. I'm not too sure I could make the calculations more transparent. I could add the explicit calculations into the header/documentation if you think it would be helpful

include/util.h
tseip: I guess I'm asking why std::round wouldn't work specifically - it should always round to the nearest integer (sometimes up, sometimes down) except in cases where that integer is unrepresentable, in which case it will round to the nearest representable int (which I don't think we should be hitting)

util.cpp
tseip: I think that can still be tested though (however this is fine)

util.cpp
tseip: No this is okay - if you have some reference for what this pairing is doing or where it's taken from you could include it in a comment (like a citation almost) which would help future coders figure out what you were doing

tseip (approved):
LGTM!

include/util.h
Victor-C-Zhang: std::round would work to give us something that could work, but we may run into an issue where it causes us to use more memory than we need to and perhaps run longer than we need to. In almost every other situation I would agree with using std::round, but since this is an area where we need to be very parsimonious with our memory usage, I'm inclined against using a round.

PR #5, opened 2020-11-12
128-bit framework

Victor-C-Zhang:
Sketches a framework for parallel 128-bit implementation and testing
- Implementation is duplicated into a `impl_128` folder
- Testing includes a new 'impl_128' folder
- 128-bit files have a `_128` extension, e.g. `graph.cpp` extends to `graph_128.cpp`
- `main.cpp` handles routing

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -1,0 +1,10 @@
+#pragma once
+
+/**
+ * Cast a double to unsigned long long with epsilon adjustment.
+ * @param d         the double to cast.
+ * @param epsilon   optional parameter representing the epsilon to use.
+ */
+unsigned long long int double_to_ull(double d, double epsilon = 0.00000001) {
+  return (unsigned long long) (d + epsilon);
+}
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -32,7 +32,27 @@ jobs:
         cd build
         cmake ../cmake_unofficial
         cmake --build .
-      
+    
+    - name: Cache boost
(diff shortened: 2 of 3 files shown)


include/util.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+/**
+ * Cast a double to unsigned long long with epsilon adjustment.
+ * @param d         the double to cast.
+ * @param epsilon   optional parameter representing the epsilon to use.
+ */
+unsigned long long int double_to_ull(double d, double epsilon = 0.00000001) {
+  return (unsigned long long) (d + epsilon);
```
tseip, line 9: Why the epsilon? If you want this to round, you should round explicitly (and also call this round) - otherwise, a straight cast will properly resolve to any ULL precisely representable by a double (this is guaranteed by the standard), and we have no reason to choose rounding slightly up for ULLs that can't be properly represented by a double.

Victor-C-Zhang: If you were to cast a straight double to int, it would work. But if we're doing calculations, the double loses precision. This has happened to [other people](https://stackoverflow.com/questions/38751915/getting-wrong-result-when-casting-from-double-to-unsigned-int), and I ran into this problem in `bucket.h` as well

Victor-C-Zhang: Of course, we run into issues with rounding up too, but I'm not sure how we can satisfy both camps when we want accurate trunctation.

tseip (approved):

PR #6, opened 2020-11-16
Added AllBucketsZeroException, added extra checks for bucket validity

kennyzzhang:
When all bucket's a b c values are zero, instead of throwing NoGoodBucketException, throw AllBucketsZeroException.

Also, added extra checks for if a bucket is good: make sure the returned index in in bounds and in the bucket.

### DIFF
--- a/test/sketch_test.cpp
+++ b/test/sketch_test.cpp
@@ -5,7 +5,7 @@
 
 TEST(SketchTestSuite, TestExceptions) {
   Sketch sketch = Sketch(10,rand());
-  ASSERT_THROW(sketch.query(), NoGoodBucketException);
+  ASSERT_THROW(sketch.query(), AllBucketsZeroException);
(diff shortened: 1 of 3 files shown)


test/sketch_test.cpp
```diff
@@ -5,7 +5,7 @@
 
 TEST(SketchTestSuite, TestExceptions) {
   Sketch sketch = Sketch(10,rand());
-  ASSERT_THROW(sketch.query(), NoGoodBucketException);
```
Victor-C-Zhang, line 8: can you write another test to check `NoGoodBucketException`?

kennyzzhang: Alright I wrote it, anything else we need?

Victor-C-Zhang (approved):

test/sketch_test.cpp
Victor-C-Zhang: looks good, lgtm

PR #7, opened 2020-11-17
Test all bucket fail

kennyzzhang:
Yeah I did too much on one branch, my bad.
- Consolidated existing sketch tests (`WHENmanySketchesAreSampledTHENallOfThemAreNonzero`, `TestAllBucketsFailProbability`) into a single test `TestSketchSample`
- Wrote `TestSketchAddition`
- Added `operator<<` for `Sketch` and `Testing_Vector` to make debugging easier
- Modified CMakeLists.txt to also find boost
- Modified CMakeLists.txt to optionally download dependencies automatically
- Modified CMake workflow to install xxHash and GTest into a cached install directory
- Modified CMake workflow to install boost into its own cached directory


CMakeLists.txt
```diff
@@ -1,26 +1,157 @@
...
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS ON)
+
 project(test)
 
-find_package(xxHash 0.7 CONFIG REQUIRED)
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  message("Adding GNU compiler flags")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -W -Wall -O3")
```
Victor-C-Zhang, line 14: what do these compiler flags do? do we want to just display warnings or treat warnings as errors as well?

.github/workflows/cmake.yml
```diff
@@ -17,45 +17,68 @@ jobs:
...
-      # Some projects don't allow in-source building, so create a separate build directory
-      # We'll use this as our working directory for all subsequent commands
-      run: cmake -E make_directory ${{runner.workspace}}/build
+    - uses: actions/cache@v2
+      id: InstalledDependencyCache
+      with:
+        path: ${{runner.workspace}}/install
+        key: InstalledDependencyCache
 
     - name: Install xxHash
```
Victor-C-Zhang, line 26: why are we installing these libraries if the cmake is set up to automatically find and downloads them?

CMakeLists.txt
kennyzzhang: They're the exact same compiler flags that we're currently using.
`-Wall` turns on most warnings
`-W` is an older name for -Wextra, turns on extra warnings
`-O3` is to optimize code a lot

In our current CMakeLists we also have `-std=c++11`, but I changed that to setting [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)

If we want to make all warnings as errors we could add -Werror, but as of right now I get a bunch of warnings when compiling our code so that's probably not a good idea

Victor-C-Zhang: If we're going to turn on warnings but ignore them, why have them on at all?

.github/workflows/cmake.yml
kennyzzhang: I'm not 100% sure how to cache the automatically downloaded packages. Without caching, every time we want to run a test it takes a few minutes to just install dependencies, most of the time being taken up by boost. The way I'm doing it now, I'm downloading the dependencies and installing them into a cached directories, so that we don't need to rebuild them every time.

include/sketch.h
```diff
@@ -48,12 +49,14 @@ class AllBucketsZeroException : public exception {
 };
 
 class MultipleQueryException : public exception {
+public:
```
Victor-C-Zhang, line 52: why does these need to be public?

CMakeLists.txt
kennyzzhang: Well we have to fix the warnings first before we make our code not able to be compiled

include/sketch.h
kennyzzhang: Otherwise calling e.what() doesn't compile, if I recall correctly

test/sketch_test.cpp
```diff
@@ -65,57 +65,125 @@ TEST(SketchTestSuite, GIVENonlyIndexZeroUpdatedTHENitWorks) {
...
-    const unsigned long num_updates = 1000;
-    Sketch sketch = Sketch(vect_size,rand());
-    Testing_Vector test_vector = Testing_Vector(vect_size,num_updates);
+/**
+ * Make sure sketch sampling works
+ */
+void test_sketch_sample(unsigned long num_sketches,
+  unsigned long vec_size, unsigned long num_updates,
+  double max_sample_fail_prob, double max_bucket_fail_prob)
+{
```
Victor-C-Zhang: nit: code style, put open brace on the end of previous line

test/sketch_test.cpp
```diff
@@ -65,57 +65,125 @@ TEST(SketchTestSuite, GIVENonlyIndexZeroUpdatedTHENitWorks) {
...
-    }
-    if (test_vector.get_entry(result.index) == 0 || test_vector.get_entry(result.index) != result.delta ){
-      ASSERT_FALSE(0) << "Failed on test " << i;
+      Update res = sketch.query();
+      //Multiple queries shouldn't happen, but if we do get here fail test
+      ASSERT_NE(res.delta, 0) << "Sample is zero";
+      ASSERT_LT(res.index, vec_size) << "Sampled index out of bounds";
+      if (res.delta != test_vec.get_entry(res.index)) {
+        //Undetected sample error
+        sample_incorrect_failures++;
```
Victor-C-Zhang, line 92: Is this one of the failure modes we should expect? If I recall, we should never return a false positive. @Abi1024 or @tenchd can either of you confirm?

test/sketch_test.cpp
Victor-C-Zhang: nit: code style

test/sketch_test.cpp
```diff
@@ -65,57 +65,125 @@ TEST(SketchTestSuite, GIVENonlyIndexZeroUpdatedTHENitWorks) {
...
-      Sketch sketch = Sketch(vec_size,rand());
-      for (int j=0;j<num_updates;++j) {
-        sketch.update(test_vec.get_update(j));
+/**
+ * Make sure sketch addition works
+ */
+void test_sketch_addition(unsigned long num_sketches,
+  unsigned long vec_size, unsigned long num_updates,
+  double max_sample_fail_prob, double max_bucket_fail_prob)
+{
```
Victor-C-Zhang: nit: code style

test/sketch_test.cpp
```diff
@@ -65,57 +65,125 @@ TEST(SketchTestSuite, GIVENonlyIndexZeroUpdatedTHENitWorks) {
...
+      sketch1.update(test_vec1.get_update(j));
+      sketch2.update(test_vec2.get_update(j));
+    }
+    Sketch sketchsum = sketch1 + sketch2;
+    try {
+      Update res = sketchsum.query();
+      ASSERT_NE(res.delta, 0) << "Sample is zero";
+      ASSERT_LT(res.index, vec_size) << "Sampled index out of bounds";
+      if (res.delta != test_vec1.get_entry(res.index) + test_vec2.get_entry(res.index)) {
+        sample_incorrect_failures++;
```
Victor-C-Zhang, line 153: we should fail immediately here. if the provided delta is not the sum of the vectors, then our addition is wrong.

.github/workflows/cmake.yml
Victor-C-Zhang: then why have a cmake step to install these libraries?

include/sketch.h
Victor-C-Zhang: we shouldn't be calling what(). It should be automatically called if an error is thrown.

CMakeLists.txt
Victor-C-Zhang: I'm confused what you're saying here. Are you saying we should fix the warnings or not?

kennyzzhang: So, if I add -Werror right now, our code will no longer be able to compile. We should eventually fix the warnings, but they are not super high priority, as long as they don't cause bugs

Victor-C-Zhang: Can you add a task for that?

test/sketch_test.cpp
kennyzzhang: Ok I know that in general we've been putting the brace on end of previous line, but in this case if I do that it's hard to tell where the function definition ends and where the body starts.

    void test_sketch_addition(unsigned long num_sketches,
      unsigned long vec_size, unsigned long num_updates,
      double max_sample_fail_prob, double max_bucket_fail_prob) {
      unsigned long all_bucket_failures = 0;
      unsigned long sample_incorrect_failures = 0;`

test/sketch_test.cpp
kennyzzhang: I'm pretty sure it's possible to get a false positive.

test/sketch_test.cpp
kennyzzhang: I'm pretty sure it's possible to get a false positive.

test/sketch_test.cpp
Victor-C-Zhang: one convention that jetbrains and a lot of companies uses is to extra-indent the overflow lines. e.g.
```
void test_sketch_addition(unsigned long num_sketches,
      unsigned long vec_size, unsigned long num_updates,
      double max_sample_fail_prob, double max_bucket_fail_prob) {
  unsigned long all_bucket_failures = 0;
  unsigned long sample_incorrect_failures = 0;
```

.github/workflows/cmake.yml
kennyzzhang: For convenience when building from scratch, i guess? I thought it was nice to be able to autodownload everything and it just works out of the box. It's just that for github workflows, since your working environment is reset each time, using autodownload would be wasting time, and I don't know of a good way to cache the results of the autodownload

include/sketch.h
kennyzzhang: According to the docs, https://en.cppreference.com/w/cpp/error/exception, what() is a "virtual public member function"

Victor-C-Zhang: Sure, but the cpp tutorial treats it as a private function http://www.cplusplus.com/doc/tutorial/exceptions/

kennyzzhang: I'm not convinced `what()` should be private.
I've been looking around and no one explicitly says it should be private, but almost all code examples I come across have it being public.
I looked into gcc's header files for standard exceptions derived from `exception`, like [`logic_error`](https://github.com/gcc-mirror/gcc/blob/700337494e1b0d5ff608e1a3c77852381e264653/libstdc%2B%2B-v3/include/std/stdexcept#L113-L149) and [`runtime_error`](https://github.com/gcc-mirror/gcc/blob/700337494e1b0d5ff608e1a3c77852381e264653/libstdc%2B%2B-v3/include/std/stdexcept#L219-L255), their overridden `what()` methods are public.
If standard exceptions have a public `what()`, why should ours be private?

I'm also not convinced we shouldn't be calling `what()`.
The tutorial you found calls `what()`.
It's also already in [our current sketch_test.cpp](https://github.com/Abi1024/StreamingGraphAlgo/blob/1fd4ee2b635f0d2197a48604c01726b538b90ff4/test/sketch_test.cpp#L85), and it got there from [one of your commits](https://github.com/Abi1024/StreamingGraphAlgo/commit/552ecf308e2785bba9f4bf70d5bec88b61048887).

Victor-C-Zhang: alright sure we can keep it

test/sketch_test.cpp
Victor-C-Zhang: I'm not convinced it's possible from the modular arithmetic part. Can you come up with a counterexample or a proof (or handwaving argument that's convincing)?

kennyzzhang: ### Hand waving:
http://dimacs.rutgers.edu/~graham//pubs/papers/l0journal.pdf: 
> We comment that there is the possibility of a false positive if one of the P1 structures erroneously reports a singleton item. This probability is polynomially small in n based on the choice of the prime p=poly(n), so we discount it.

https://www.juanlopes.net/poly18/poly18-juan-lopes.pdf:
> Theorem 1.
If a is 1-sparse, then b2 is congruent to b0z^(b1/b0) mod p. Otherwise, b2 is not congruent to b0z^(b1/b0) mod p with probability at least 1−n/p.

### Counterexamples:
Prime=101, r=29
|Index:|1|2|3|4|5|6|7|8|9|10|
|-|-|-|-|-|-|-|-|-|-|-|
|Vector:|-5|-5|0|0|0|-5|2|-2|1|0|
|Bucket hash:|1|1|1|0|0|1|0|0|0|1|

a=-5-5-5=-15
b=-5\*1-5\*2-5\*6=-45
c=(-5\*29\*\*1-5\*29\*\*2-5\*29\*\*6)%101=88
ar^(b/a)=(-15\*29\*\*3)%101=88

Prime=101, r=85
|Index:|1|2|3|4|5|6|7|8|9|10|
|-|-|-|-|-|-|-|-|-|-|-|
|Vector:|2|-3|0|-7|4|0|0|-1|-2|-5|
|Bucket hash:|1|1|1|1|1|0|1|0|0|0|

a=2-3-7+4=-4
b=2\*1-3\*2-7\*4+4\*5=-12
c=(2\*85\*\*1-3\*85\*\*2-7\*85\*\*4+4\*85\*\*5)%101=22
ar^(b/a)=(-4\*85\*\*3)%101=22

Prime=101, r=70
|Index:|1|2|3|4|5|6|7|8|9|10|
|-|-|-|-|-|-|-|-|-|-|-|
|Vector:|2|0|0|1|0|3|-8|0|-3|-2|
|Bucket hash:|1|1|0|1|1|1|0|0|0|0|

a=2+1+3=6
b=2\*1+1\*4+3\*6=24
c=(2\*70\*\*1+1\*70\*\*4+3\*70\*\*6)%101=64
ar^(b/a)=(6\*70\*\*4)%101=64

There's an entirely separate problem I noticed, that if r=0 or 1, the entire c test is worthless. We should probably do something about that.
I also noticed a and b are different in between our jamboard and our code, not a huge issue but something to note.

Victor-C-Zhang: I see, this is interesting. Did you fix the r problem in this PR? I think we can just put r \in [2, p) if that's the only issue

Victor-C-Zhang (approved):
LGTM!

PR #8, opened 2020-12-14
Made tests check against theoretical probability, added variety of tests

kennyzzhang:


test/sketch_test.cpp
```diff
@@ -71,8 +71,20 @@ TEST(SketchTestSuite, GIVENonlyIndexZeroUpdatedTHENitWorks) {
  * Make sure sketch sampling works
  */
 void test_sketch_sample(unsigned long num_sketches,
-    unsigned long vec_size, unsigned long num_updates,
-    double max_sample_fail_prob, double max_bucket_fail_prob) {
+    unsigned long vec_size, unsigned long num_updates) {
+  /* We expect the all bucket failure probability to be 1/(n^c)
+   * Currently c is hardcoded to 1
+   */
+  double max_bucket_fail_prob = 1./vec_size;
```
Victor-C-Zhang, line 78: Can you declare a `c` variable somewhere in the codebase and pull that here? That way, we hardcode `c` once, and changing that value will automatically update `sketch.cpp` and all of these tests that rely on `c`.

test/sketch_test.cpp
```diff
@@ -109,27 +121,47 @@ void test_sketch_sample(unsigned long num_sketches,
...
  * Make sure sketch addition works
  */
 void test_sketch_addition(unsigned long num_sketches,
-    unsigned long vec_size, unsigned long num_updates,
-    double max_sample_fail_prob, double max_bucket_fail_prob) {
+    unsigned long vec_size, unsigned long num_updates) {
+  /* We expect the all bucket failure probability to be 1/(n^c)
+   * Currently c is hardcoded to 1
+   */
+  double max_bucket_fail_prob = 1./vec_size;
```
Victor-C-Zhang, line 155: use `c` here as well

PR #9, opened 2020-12-22
Trim bucket

Victor-C-Zhang:
Trim buckets to only include fields a,b,c.
Created separate extended implementation using custom `int127` type built on top of `boost::multiprecision::uint128_t` to handle signed integer computations. Modified CI workflows to run tests against both native and extended implementations.


include/prime_generator.h
```diff
@@ -1,28 +1,40 @@
 #pragma once
```
Abi1024, line 1: I've noticed you switched from the AKS primality test to the naive square root test. I would guess that this is to accommodate the extended integer types. Is that correct? If so, I think that's alright if we are generating one prime per sketch, and don't have too many sketches. Ultimately, the ideal solution here would be to use a version of AKS that does support extended types.

include/sketch.h
```diff
@@ -13,10 +19,12 @@ using namespace std;
 class Sketch {
   const long seed;
   const uint64_t n;
-  std::vector<Bucket> buckets;
-  const long long int random_prime;
+  std::vector<Bucket_Boruvka> buckets;
+  const uint128_t large_prime;
   bool already_quered = false;
 
+  FRIEND_TEST(SketchTestSuite, TestExceptions);
```
Abi1024, line 26: I'm still new to gtest, so I assume this is a gtest construct that links this code to the testing suite?

l0_sampling/sketch.cpp
```diff
@@ -1,52 +1,68 @@
...
+      XXH64_hash_t bucket_seed = XXH64(&bucket_id, sizeof(bucket_id), seed);
+      int128_t r = 2 +  bucket_seed % (large_prime - 3);
+      if (buckets[bucket_id].contains(update.index+1, bucket_seed, 1 << j)){
+        buckets[bucket_id].a += update.delta;
+        buckets[bucket_id].b += update.delta*(update.index+1); // deals with updates whose indices are 0
+        buckets[bucket_id].c = static_cast<uint128_t>(
+              (static_cast<int128_t>(buckets[bucket_id].c)
+              + static_cast<int128_t>(large_prime)
+              + (update.delta*PrimeGenerator::power(r,(uint128_t) update
+              .index+1, large_prime) % static_cast<int128_t>(large_prime)))
```
Abi1024, line 31: Just to make sure I'm not going crazy, what is the +1 for here? I presume it is to deal with the 0 indexing issue we had before?

l0_sampling/sketch.cpp
```diff
@@ -1,52 +1,68 @@
...
+  const unsigned long long int num_guesses = guess_gen(n);
+  for (unsigned i = 0; i < num_buckets; ++i) {
+    for (unsigned j = 0; j < num_guesses; ++j) {
+      Bucket_Boruvka& b = buckets[i*num_guesses+j];
+      if (b.a != 0 || b.b != 0 || b.c != 0) {
+        all_buckets_zero = false;
+      }
+      unsigned bucket_id = i * num_guesses + j;
+      XXH64_hash_t bucket_seed = XXH64(&bucket_id, sizeof(bucket_id), seed);
+      uint128_t r = 2 + bucket_seed % (large_prime - 3);
```
Abi1024, line 55: I get that r should not be {-1,0,1}, but why we do we have (large prime - 3).

include/sketch.h
kennyzzhang: It makes a test a friend of a class. To generate a vector where all buckets fail, we need to know which indices each bucket contains, which is private.

l0_sampling/sketch.cpp
kennyzzhang: You are correct, the +1 is to deal with the 0 index issue we had before.

l0_sampling/sketch.cpp
kennyzzhang: r should not be {-1,0,1}, and since we are working with mod large_prime, r should not be {0,1,large_prime-1}.
This means we want an integer in the range [2,large_prime-2].

rand%m gives an integer in the range [0,m-1].
bucket_seed % (large_prime - 3) gives an integer in the range [0,large_prime - 4].
2 + bucket_seed % (large_prime - 3) gives an integer in the range [2,large_prime - 2], which is what we're looking for.

include/prime_generator.h
kennyzzhang: Hmm... I'm looking over our code and I don't think we ever used AKS. From what I can tell, the IsPrime function has always been the naive square root test and all the other things in aks.h/cpp (LargestPrimeFactor, Powm, Inv, CPolyMod) have never been used. The files aks.h and aks.cpp were created in commit 7ef0632aa6536df1b01afe50ae492330f21011a9, and they were using the naive primality test back then.

Victor-C-Zhang: We haven't been using AKS to do primality testing. The `IsPrime` function we'd been using is essentially a brute-force prime checker.

test/int127_test.cpp
```diff
@@ -0,0 +1,70 @@
+#include <gtest/gtest.h>
```
Abi1024, line 1: I think it's important to have comments explaining what each test does. Just a line or two should suffice.

Abi1024 (approved):
Good work on the implementations. I think having comments in the tests are very important though, so try to add those in when you get a chance.

PR #12, opened 2021-01-01
Correctness graph tests

Victor-C-Zhang:
Implements points 1 and 2 in https://github.com/Abi1024/StreamingGraphAlgo/wiki/Connected-Components-Testing#correctness-tests

**Note:** the diff isn't actually 5700 lines! Roughly 5200 of these lines comes from a sample input file in `test/res`


include/graph.h
```diff
@@ -1,23 +1,37 @@
...
+ */
 class Graph{
   const uint64_t num_nodes;
   bool update_locked = false;
   // a set containing one "representative" from each supernode
   set<Node>* representatives;
   Supernode** supernodes;
 
+#ifdef VERIFY_SAMPLES_F
+  std::string cum_in = "./cum_sample.txt";
```
etwest: I think this should be moved to the ifdef on line 49. As far as I can see cum_in isn't used before that point. Fewer ifdefs = simpler.

test/util/graph_gen.h
```diff
@@ -0,0 +1,29 @@
...
+  long n;            // number of nodes
+  double p;          // prob of edge between nodes
+  long m;           // number of edges
+  bool graph_model;     // G(n,p) -- 0, G(n,m) -- 1
+  int max_appearances;  // the maximum number of times an edge can show up
+                            // in the stream
+  std::string out_file; // file to write stream
+  std::string cum_out_file; // file to write cum graph
+  genSet(long n, double p, long m, bool graph_model, int max_appearances,
+         std::string out_file, std::string cum_out_file)
```
etwest, line 16: You mention elsewhere that sample.txt and cum_sample.txt are the default values for out_file and cum_out_file. It might be better to set those here to avoid needing to specify elsewhere.

test/graph_test.cpp
```diff
@@ -1,26 +1,88 @@
...
   }
+  g.set_cum_in(curr_dir + "/res/multiples_graph_1024.txt");
   g.connected_components();
   ASSERT_THROW(g.update({{1,2}, INSERT}), UpdateLockedException);
   ASSERT_THROW(g.update({{1,2}, DELETE}), UpdateLockedException);
 }
+
+TEST(GraphTestSuite, TestRandomGraphGeneration) {
+  generate_stream();
+  GraphVerifier graphVerifier {"./cum_sample.txt"};
```
etwest: I'm not 100% sure about this syntax but isn't cum_sample a default arg. Do we need to specify.

include/graph.h
```diff
@@ -1,23 +1,37 @@
...
 
+#ifdef VERIFY_SAMPLES_F
+#include "../test/util/deterministic.h"
+#endif
+
 enum UpdateType {
-  INSERT,
-  DELETE,
+  INSERT = 0,
+  DELETE = 1,
```
etwest, line 15: Is this necessary? Aren't these just the default values for an enum. If you have a reason feel free to keep the change.

test/util/deterministic.cpp
```diff
@@ -0,0 +1,108 @@
+#include <fstream>
```
etwest: Why the name deterministic.cpp/h? Seems like graph_verifier.cpp/h might fit the content better?

test/util/graph_gen.h
Victor-C-Zhang: They're set as default params in `generate_stream()`. I'd like to keep them there, since technically it's a standard that we're setting, not one set by `GTgraph`

test/graph_test.cpp
Victor-C-Zhang: You're right, it's not needed

include/graph.h
Victor-C-Zhang: I'd like to keep them for reference since these are the values that show up in the `sample.txt` file for `INSERT` and `DELETE`

etwest: Okay then I have no problem with this :)

etwest (approved):
Looks good to go.

PR #13, opened 2021-01-18
bug fix for prime generation.

etwest:
I think I found a bug where prime generation is returning a composite number rather than a prime number. Not sure exactly what the implications are of this issue but I'm sure it's not great.

Let me know if I'm missing something but I think the following is the solution.

prime_generator.h -> change line 37 to while(!isPrime())
prime_generator_native.h -> change line 38 to while(!isPrime())

I built a quick testing script and got the following results:
prime generated from 5 is 9
prime generated from 10 is 15
prime generated from 23 is 25

This pull request should fix this issue

### DIFF
--- a/include/prime_generator.h
+++ b/include/prime_generator.h
@@ -34,7 +34,7 @@ namespace PrimeGenerator{
     if (n % 2 == 0){
       n++;
     }
-    while(IsPrime(n)){
+    while(!IsPrime(n)){
       n += 2;
     }
     return n;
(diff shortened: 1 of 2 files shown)


Victor-C-Zhang:
The `IsPrime` function returns as intended though, right?

etwest:
> The `IsPrime` function returns as intended though, right?

Yes. IsPrime correctly identifies if a given number is prime.

Victor-C-Zhang (approved):

PR #14, opened 2021-02-08
Batched sketch updates/refactoring

kennyzzhang:
I refactored our code a bunch as well. It used to be a pretty tangled mess of repeated code in all sorts of places, two separate implementations just for a change in types. Now there's a types.h, which is a mess of \#ifdefs, but it makes swapping between native implementation and boost::multiprecision way easier.


include/types.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#ifdef USE_NATIVE_F
+//Native type includes
+#ifndef __SIZEOF_INT128__
+#include <boost/multiprecision/cpp_int.hpp>
+#endif
```
Victor-C-Zhang: can you comment which ifdef this endif corresponds to? and similarly for other endifs

include/types.h
```diff
@@ -0,0 +1,67 @@
...
+#ifdef __SIZEOF_INT128__
+typedef __int128 bucket_prod_t;
+typedef unsigned __int128 ubucket_prod_t;
+#else
+typedef boost::multiprecision::int128_t bucket_prod_t;
+typedef boost::multiprecision::uint128_t ubucket_prod_t;
+#endif
+#else
+//128 bit types
+typedef uint64_t node_t; //Max graph vertices is 92682
```
Victor-C-Zhang: is this true?

include/types.h
```diff
@@ -0,0 +1,67 @@
...
+typedef boost::multiprecision::uint128_t ubucket_t;
+#endif
+typedef boost::multiprecision::int256_t bucket_prod_t;
+typedef boost::multiprecision::uint256_t ubucket_prod_t;
+#endif
+
+#ifdef __SIZEOF_INT128__
+//overload operator<< for __int128 and __uint128
+inline std::ostream& operator<<(std::ostream& os, const __int128& n) {
+	//TODO: Endianness?
```
Victor-C-Zhang: comment which type of endianness you've written for

include/bucket.h
```diff
@@ -1,37 +1,34 @@
...
+
+/**
+ * Represents a bucket in a sketch.
+ */
+struct Bucket_Boruvka {
+  bucket_t a = 0;
+  bucket_t b = 0;
+  ubucket_t c = 0;
+
+  static XXH64_hash_t gen_bucket_seed(const unsigned bucket_id, long seed);
```
Victor-C-Zhang: consider making this inline

include/bucket.h
```diff
@@ -1,37 +1,34 @@
...
+  static ubucket_t gen_r(const XXH64_hash_t& bucket_seed, const ubucket_t& large_prime);
+  /**
+   * Checks whether the hash associated with the Bucket hashes the index to 0.
+   * @param index
+   * @param bucket_seed
+   * @param guess_nonzero
+   * @return true if the index is NOT hashed to zero.
+   */
+  bool contains(const vec_t& index, const XXH64_hash_t& bucket_seed, const vec_t& guess_nonzero) const;
+  bool is_good(const vec_t& n, const ubucket_t& large_prime, const XXH64_hash_t& bucket_seed, const ubucket_t& r, const vec_t& guess_nonzero) const;
```
Victor-C-Zhang, line 58: can you document these functions?

Victor-C-Zhang (commented):
Now that we have types file, we don't need `x_native` files anymore

include/graph.h
```diff
@@ -35,6 +37,8 @@ class Graph{
   ~Graph();
   void update(GraphUpdate upd);
 
+  void batch_update(uint64_t src, const std::vector<std::pair<uint64_t, int8_t>>& edges);
```
Victor-C-Zhang, line 45: documentation for this as well

include/supernode.h
```diff
@@ -48,11 +46,13 @@ class Supernode {
    * BEFORE Boruvka starts.
    */
   void update(pair<Edge, int> update);
+
+  void batch_update(const std::vector<Update>& updates);
```
Victor-C-Zhang, line 54: document as well

include/types.h
Victor-C-Zhang: if we allow 256 bit p^2, this should be larger right?

Victor-C-Zhang (approved):

PR #17, opened 2021-03-26
Hashxor

kennyzzhang:


include/bucket.h
```diff
@@ -2,76 +2,93 @@
...
+inline vec_t Bucket_Boruvka::col_index_hash(const unsigned bucket_col, const vec_t& update_idx, const long sketch_seed) {
+  struct {
+    unsigned bucket_col;
+    vec_t update_idx;
+  } __attribute__((packed)) buf = {bucket_col, update_idx};
+  return
+#ifdef USE_NATIVE_F
+  XXH32
+#else
+  XXH3_64bits_withSeed
```
Victor-C-Zhang, line 67: what's the difference between XXH32 and XXH3?

include/sketch.h
```diff
@@ -4,13 +4,11 @@
...
 #include <vector>
 #include "bucket.h"
-#include "prime_generator.h"
 #include "types.h"
-#include "update.h"
 #include "util.h"
 #include <gtest/gtest_prod.h>
 
-#define bucket_gen(x) double_to_ull(log2(x)+1)
+#define bucket_gen(x, c) double_to_ull((c)*(log2(x)+1))
```
Victor-C-Zhang, line 11: I'm not sure this is the way we want to do this. We want to convert `log2(x)` to an integer, then multiply that by `c` right?

include/sketch.h
```diff
@@ -21,39 +19,37 @@
 class Sketch {
   const long seed;
   const vec_t n;
+  const double num_bucket_factor;
```
Victor-C-Zhang: can you document these fields?

l0_sampling/sketch.cpp
```diff
@@ -56,32 +60,30 @@ Update Sketch::query() {
...
+    sketch1.buckets[i].a ^= sketch2.buckets[i].a;
+    sketch1.buckets[i].c ^= sketch2.buckets[i].c;
   }
   sketch1.already_quered = sketch1.already_quered || sketch2.already_quered;
   return sketch1;
 }
 
-
+//Probably doesn't work anymore
+/*
```
Victor-C-Zhang: we don't ever use this, so I think we can just get rid of it

test/sketch_test.cpp
```diff
@@ -225,35 +226,42 @@ void test_sketch_large(unsigned long vec_size, unsigned long num_updates) {
 }
 
 TEST(SketchTestSuite, TestSketchLarge) {
-  test_sketch_large(10000000, 1000000);
+  constexpr uint64_t upper_bound =
+#ifdef USE_NATIVE_F
+      1000000000ULL
```
Victor-C-Zhang: `1e9`

test/sketch_test.cpp
```diff
@@ -225,35 +226,42 @@ void test_sketch_large(unsigned long vec_size, unsigned long num_updates) {
...
+#else
+      1000000000000000000ULL
```
Victor-C-Zhang: `1e18`

Victor-C-Zhang (changes requested):
We can get rid of the prime-related files, as well as the `GTgraph`-related binaries in `/test/util/`

include/sketch.h
kennyzzhang: I don't think the order of the conversion matters that much. We're toying with c=.5 and other non-integer factors, so we may as well do just the integer conversion afterwards.

include/bucket.h
kennyzzhang: Here's the benchmark table from XXH's page:
| Hash Name     | Width | Bandwidth (GB/s) | Small Data Velocity | Quality | Comment |
| ---------     | ----- | ---------------- | ----- | --- | --- |
| __XXH3__ (SSE2) |  64 | 31.5 GB/s        | 133.1 | 10
| __XXH128__ (SSE2) | 128 | 29.6 GB/s      | 118.1 | 10
| _RAM sequential read_ | N/A | 28.0 GB/s  |   N/A | N/A | _for reference_
| City64        |    64 | 22.0 GB/s        |  76.6 | 10
| T1ha2         |    64 | 22.0 GB/s        |  99.0 |  9 | Slightly worse [collisions]
| City128       |   128 | 21.7 GB/s        |  57.7 | 10
| __XXH64__     |    64 | 19.4 GB/s        |  71.0 | 10
| SpookyHash    |    64 | 19.3 GB/s        |  53.2 | 10
| Mum           |    64 | 18.0 GB/s        |  67.0 |  9 | Slightly worse [collisions]
| __XXH32__     |    32 |  9.7 GB/s        |  71.9 | 10
| City32        |    32 |  9.1 GB/s        |  66.0 | 10
| Murmur3       |    32 |  3.9 GB/s        |  56.1 | 10
| SipHash       |    64 |  3.0 GB/s        |  43.2 | 10
| FNV64         |    64 |  1.2 GB/s        |  62.7 |  5 | Poor avalanche properties
| Blake2        |   256 |  1.1 GB/s        |   5.1 | 10 | Cryptographic
| SHA1          |   160 |  0.8 GB/s        |   5.6 | 10 | Cryptographic but broken
| MD5           |   128 |  0.6 GB/s        |   7.8 | 10 | Cryptographic but broken

So for "native" we're using 32 bit integers, so may as well use the 32 bit hash XXH32.
For 64 bit integers, we were using XXH64, but it turns out XXH3 is just better in every way.

include/sketch.h
Victor-C-Zhang: Alright, sounds good

Victor-C-Zhang (approved):

PR #22, opened 2021-04-22
OpenMP for stream ingestion

Victor-C-Zhang:
Use OpenMP to parallelize `Supernode.batch_update`


test/experiment/parallelism.cpp
```diff
@@ -0,0 +1,82 @@
...
+    free(prime);
+  }
+
+  void SetUp() override {}
+  void TearDown() override {}
+};
+
+vector<Edge>* EXPR_Parallelism::graph_edges;
+vector<Edge>* EXPR_Parallelism::odd_graph_edges;
+bool* EXPR_Parallelism::prime;
```
kennyzzhang: I can't find anywhere this is referenced, do we really need this test fixture?

Victor-C-Zhang: probably not, I'll get rid of it

kennyzzhang (approved):

PR #28, opened 2021-05-14
32 bit for bucket checksum

kennyzzhang:


l0_sampling/sketch.cpp
```diff
@@ -61,44 +60,44 @@ vec_t Sketch::query() {
 }
 
 Sketch operator+ (const Sketch &sketch1, const Sketch &sketch2){
```
Victor-C-Zhang: Why don't we need to check seed is the same?

kennyzzhang: `operator+` calls `operator+=`, `operator+=` checks that seed is the same. I changed this because felt redundant to have two copies of the same code

Victor-C-Zhang: I see, that makes sense. As a followup, do we really need `operator+`? I don't think we use it anywhere in prod

kennyzzhang: I guess we don't, should i remove it?

Victor-C-Zhang: In that case yeah.

Victor-C-Zhang (approved):

PR #29, opened 2021-05-21
Buffer Tree and Node Level Parallelism

etwest:
This is almost ready for integration. Just need to test it on the lab machine.


etwest:
Also, most of the added lines here are some test files. We don't need to include those but it might be helpful to have?

etwest:
I have a counter in the graph.cpp which tracks the total number of updates to the data structure that have been processed. I have it marked for removal but it's actually fairly useful. Do we want to keep it?

Victor-C-Zhang (commented):
I think it's ok if we don't include the massive graph files. I'm not sure it does too much for us and we can always re-generate the files the `efficient_gen` executable.

etwest:
Removed the test files. Should be ready for merging. Please mark as reviewed.

CMakeLists.txt
```diff
@@ -1,12 +1,16 @@
 #TODO: write FindxxHash for libxxhash-dev compatibility?
 #Cmake for xxHash is unofficial, so libxxhash-dev doesn't have xxHashConfig.cmake
 
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.15)
```
Victor-C-Zhang, line 4: why?

CMakeLists.txt
```diff
@@ -182,10 +207,15 @@ add_executable(tests
   test/util/graph_verifier_test.cpp
   graph.cpp
   supernode.cpp
+  GraphWorker.cpp
```
Victor-C-Zhang: nit: can you change the name to match our other naming standard? lowercase_with_underscores

CMakeLists.txt
etwest: Needed for openmp
3.1 doesn't support the cmake commands used

CMakeLists.txt
etwest: Sure

GraphWorker.cpp
```diff
@@ -0,0 +1,78 @@
...
+				printf("Number of groups = %i\n", num_groups);
+			}
+			if(line.substr(0, line.find('=')) == "group_size") {
+				group_size = std::stoi(line.substr(line.find('=')+1));
+				printf("Size of groups = %i\n", group_size);
+			}
+		}
+			
+	} else {
+		printf("WARNING: Could not open thread configuration file!\n");
```
Victor-C-Zhang: Add default configuration notification

GraphWorker.cpp
```diff
@@ -0,0 +1,78 @@
...
+	}
+	workers = (GraphWorker **) calloc(num_groups, sizeof(GraphWorker *));
```
Victor-C-Zhang: why calloc?

graph.cpp
```diff
@@ -15,6 +18,13 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
     supernodes[i] = new Supernode(num_nodes,seed);
     parent[i] = i;
   }
+  num_updates = 0; // REMOVE this later
```
Victor-C-Zhang, line 21: do we need this?

GraphWorker.cpp
```diff
@@ -0,0 +1,78 @@
...
+	}
+	delete workers;
+}
+
+GraphWorker::GraphWorker(int _id, Graph *_graph, BufferTree *_bf) :
+  id(_id), graph(_graph), bf(_bf), thr(startWorker, this) {
+}
+
+GraphWorker::~GraphWorker() {
+	thr.join();
```
Victor-C-Zhang: Have you thought about detaching the thread and letting it run until we set the `shutdown` flag?

GraphWorker.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+void GraphWorker::doWork() {
+	while(true) {
+		std::unique_lock<std::mutex> queue_unique(bf->queue_lock);
+		bf->queue_cond.wait(queue_unique, [this]{return (bf->work_queue.empty() == false || shutdown);});
+
+		if (bf->work_queue.empty() == false) {
```
Victor-C-Zhang: `!empty()`

GraphWorker.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+GraphWorker::~GraphWorker() {
+	thr.join();
+}
+
+void GraphWorker::doWork() {
+	while(true) {
+		std::unique_lock<std::mutex> queue_unique(bf->queue_lock);
+		bf->queue_cond.wait(queue_unique, [this]{return (bf->work_queue.empty() == false || shutdown);});
```
Victor-C-Zhang: `!empty()`

GraphWorker.cpp
```diff
@@ -0,0 +1,78 @@
...
+			queue_unique.unlock();  // unlock
+			// printf("Worker %d handling updates for node %llu\n", id, node);
+			std::pair<Node, std::vector<Node>> data = bf->get_data(task);
+			graph->batch_update(data.first, data.second);
+		}
+		else queue_unique.unlock(); // unlock
+
+		// doesn't really matter if this part isn't thread safe I believe
+		// only reading and never should read empty when there is more to insert
+		if (shutdown && bf->work_queue.empty()) {// if recieved shutdown and there's no more work to do
```
Victor-C-Zhang: will we have some problems with conditional waiting if this isn't synchronized?

graph.cpp
```diff
@@ -2,6 +2,9 @@
 #include <iostream>
 #include "include/graph.h"
 
+#include <buffer_tree.h>
```
Victor-C-Zhang, line 5: < nice >

graph.cpp
```diff
@@ -50,16 +58,23 @@ void Graph::batch_update(uint64_t src, const std::vector<uint64_t>& edges) {
       updates.push_back(static_cast<vec_t>(
           nondirectional_non_self_edge_pairing_fn(edge, src)));
     }
+    num_updates += 1; // REMOVE this later
```
Victor-C-Zhang, line 61: same with this

GraphWorker.cpp
etwest: What do you mean? you want me to change the text?

graph.cpp
etwest: Maybe, maybe not. It's been helpful to me to have it but it's definitely not strictly necessary

GraphWorker.cpp
etwest: No. We want to wait on the thread finishing. The shutdown message is not an immediate stop rather it just tells the thread to exit when there is no more work to do.

GraphWorker.cpp
etwest: No particular reason. Malloc would work as well

include/graph.h
```diff
@@ -5,13 +5,17 @@
 #include <set>
 #include <fstream>
 #include "supernode.h"
+#include <atomic>  // REMOVE LATER
```
Victor-C-Zhang, line 8: hm

CMakeLists.txt
Victor-C-Zhang: alright

GraphWorker.cpp
etwest: I don't believe so but perhaps I'm wrong.
The logic is that checking if the work_queue is empty is not an unsafe operation and that the work queue is full with all of the updates we need to process before shutdown is true.

graph_worker.cpp
```diff
@@ -0,0 +1,78 @@
...
+	}
+	delete workers;
+}
+
+GraphWorker::GraphWorker(int _id, Graph *_graph, BufferTree *_bf) :
+  id(_id), graph(_graph), bf(_bf), thr(startWorker, this) {
+}
+
+GraphWorker::~GraphWorker() {
+	thr.join();
```
Victor-C-Zhang, line 56: have you considered detaching the thread instead of waiting to join?

run_experiments.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/bash
```
Victor-C-Zhang: put this into the `test/experiment` folder

include/graph.h
etwest: Yeah so this is for that counter that I have... if you think it should go I'm happy to remove it.

graph_worker.cpp
etwest: No. We want to wait on the thread finishing. The shutdown message is not an immediate stop rather it just tells the thread to exit when there is no more work to do.

test/experiment/BCHMK_Graph.cpp
```diff
@@ -0,0 +1,42 @@
...
+  Graph g{num_nodes};
+  printf("Insertions\n");
+  printf("Progress:                    | 0%%\r"); fflush(stdout);
+  clock_t start = clock();
+  while (m--) {
+    if ((total - m) % (int)(total * .05) == 0) {
+      clock_t diff = clock() - start;
+      float num_seconds = diff / CLOCKS_PER_SEC;
+      int percent = (total - m) / (total * .05);
+      printf("Progress:%s%s", std::string(percent, '=').c_str(), std::string(20 - percent, ' ').c_str());
```
Victor-C-Zhang, line 30: this is pretty cool

run_experiments.sh
etwest: Done

test/experiment/BCHMK_Graph.cpp
etwest: Yeah I thought so too :) I had fun with it.

test/util/graph_verifier.h
```diff
@@ -51,10 +51,24 @@ class GraphVerifier {
...
+//   }
+// };
+
+// class BadEdgeException2 : public exception {
+//   virtual const char* what() const throw() {
+//     return "The edge is not in the cut of the sample! (2)";
+//   }
+// };
+
+
```
Victor-C-Zhang: no need for this since we have something similar in `graph_verifier.cpp`

graph.cpp
Victor-C-Zhang: What would be its purpose in production?

graph_worker.cpp
Victor-C-Zhang: Ah ok, and we're doing that implicitly with the delete call. Maybe document this somewhere.

GraphWorker.cpp
Victor-C-Zhang: Yes, since we default to 1 and 1 our users (probably me) should be notified of that.

GraphWorker.cpp
Victor-C-Zhang: Ok, I see. But I think a similar problem remains because the condition variable isn't going to be notified when the queue is empty. So if we have a number of workers in this while loop after shutdown has been called, all but one of them will be stuck waiting for a spurious wakeup to exit the loop.

Victor-C-Zhang: This is probably not too big of an issue in the grand scheme of things, but it would be nice if there was an easy fix for this

graph_worker.cpp
etwest: Will do

test/util/graph_verifier.h
etwest: Ok

GraphWorker.cpp
etwest: Actually I think this is okay because the wait condition includes shutdown. So if shutdown is true then we won't ever wait on the condition variable.

graph_worker.cpp
```diff
@@ -0,0 +1,80 @@
...
+				printf("Number of groups = %i\n", num_groups);
+			}
+			if(line.substr(0, line.find('=')) == "group_size") {
+				group_size = std::stoi(line.substr(line.find('=')+1));
+				printf("Size of groups = %i\n", group_size);
+			}
+		}
+			
+	} else {
+		printf("WARNING: Could not open thread configuration file!\n");
```
Victor-C-Zhang, line 30: Can you add a note telling us what the config values default to?

GraphWorker.cpp
Victor-C-Zhang: Is that how condition variables work? I was under the impression the condition variable only checks the predicate after it's woken up.

graph.cpp
etwest: Might not have one. Depends if we think it's useful now. Although perhaps it would be nice to be able to verify that the total number of updates to the graph is what we expected it would be.

graph_worker.cpp
etwest: Ha yes just did

graph.cpp
Victor-C-Zhang: Sure, I guess we can keep it for now.

Victor-C-Zhang (approved):

PR #30, opened 2021-05-25
Functions to write things explicitly to files

Victor-C-Zhang:
Functions to write and re-read `Graph`, `Supernode`, and `Sketch` objects to secondary storage.

### DIFF
--- a/graph.cpp
+++ b/graph.cpp
@@ -40,6 +40,7 @@ Graph::Graph(std::ifstream& file) {
     supernodes[i] = new Supernode(num_nodes, file);
     parent[i] = i;
   }
+  is_copy = true;
 }
 
 Graph::~Graph() {
@@ -48,7 +49,7 @@ Graph::~Graph() {
   delete[] supernodes;
   delete[] parent;
   delete representatives;
-  delete bf;
+  if (!is_copy) delete bf;
 }
 
 void Graph::update(GraphUpdate upd) {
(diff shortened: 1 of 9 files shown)


graph.cpp
```diff
@@ -27,13 +27,29 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
   GraphWorker::startWorkers(this, bf);
 }
 
+Graph::Graph(std::ifstream& file) {
```
etwest, line 30: So the way we're thinking about using this in multiple CC is to reset an existing graph's sketches using the data on disk. How would we accomplish that using this setup?

graph.cpp
```diff
@@ -27,13 +27,29 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
...
+#endif
+  representatives = new set<Node>();
+  supernodes = new Supernode*[num_nodes];
+  parent = new Node[num_nodes];
+  for (Node i=0;i<num_nodes;++i) {
+    representatives->insert(i);
+    supernodes[i] = new Supernode(num_nodes, file);
+    parent[i] = i;
+  }
+  is_copy = true;
```
etwest, line 43: Same with this variable. There are a number of places where we do different things for copies. How would that fit into what we want to do with this.

graph.cpp
Victor-C-Zhang: Constructing from a filestream saves only the supernode and metadata fields. It wasn't made with the intention of being able to be a functional ingestion machine after construction. That's why I use the term "reheated" like you would yesterday's dinner. I'll rethink this whole process in light of our new need for a snapshot-based serialization.

Victor-C-Zhang:
Save this for later

PR #32, opened 2021-05-26
Buffer Tree Configuration

etwest:
Fixed an issue where the location of the buffer tree data could not be defined by the user. Additionally, rearchitected the code for defining a configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@
 
 #Doxygen docs
 /docs/
+
+# Our configuration file
+streaming.conf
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -47,3 +47,6 @@
 
 #Doxygen docs
 /docs/
+
+# Our configuration file
+streaming.conf
```
Victor-C-Zhang, line 52: We're "ignoring" this file, but it's still showing up in the diff. Is this intentional?

etwest: Yes. I want the initial configuration to be added to the repo but any changes folks make should be ignored in commits.

Victor-C-Zhang (approved):

PR #33, opened 2021-05-30
Integration of Changes to Buffer Tree

etwest:
We will first need to merge the pull request in the buffer-tree. Then commit a change to this pull request which changes from the circular_queue branch to the main branch of the buffer-tree.
(and ideally do that fast enough that no one builds the system in between)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,7 @@ ExternalProject_Add(
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "circular_queue"
+  GIT_TAG         "main"
 
   BUILD_ALWAYS    OFF
   INSTALL_DIR     ${CMAKE_CURRENT_BINARY_DIR}/BufferTree/prefix
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,7 @@ ExternalProject_Add(
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "circular_queue"
 
   BUILD_ALWAYS    OFF
   INSTALL_DIR     ${CMAKE_CURRENT_BINARY_DIR}/BufferTree/prefix
--- a/graph_worker.cpp
+++ b/graph_worker.cpp
@@ -11,13 +11,12 @@ int GraphWorker::group_size = 1;
 GraphWorker **GraphWorker::workers;
 
 void GraphWorker::startWorkers(Graph *_graph, BufferTree *_bf) {
-	
+	shutdown = false;
+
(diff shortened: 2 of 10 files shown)


CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ ExternalProject_Add(
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "circular_queue"
```
Victor-C-Zhang: remember to change this once we merge to main

graph_worker.cpp
```diff
@@ -33,31 +32,24 @@ GraphWorker::GraphWorker(int _id, Graph *_graph, BufferTree *_bf) :
...
+		// call get_data which will handle waiting on the queue
+		// and will enforce locking.
+		bool valid = bf->get_data(data);
 
-		// doesn't really matter if this part isn't thread safe I believe
-		// only reading and never should read empty when there is more to insert
-		if (shutdown && bf->work_queue.empty()) {// if recieved shutdown and there's no more work to do
-			printf("Thread %i done and exiting\n", id);
+		if (valid)
+			graph->batch_update(data.first, data.second);
```
Victor-C-Zhang, line 51: wait for implicit/explicit sync impl

etwest: What?

CMakeLists.txt
etwest: Yep!

graph_worker.cpp
Victor-C-Zhang: since `batch_update` isn't synchronized, multiple workers could try to access a sketch concurrently. we need to sync access, either implicitly (within `sketch::operator+=`) or explicitly.

etwest: Oh yes. Absolutely, however this change (the circular queue) isn't what will make that necessary. The basement nodes are what will do that.

etwest:
This most recent commit will fail until I merge the pull request on the buffer tree side. I'd like to have an approving review before that happens so we can do this quickly.

Victor-C-Zhang:
This code relies on sync to run without error. Before we can merge this we need to implement sync. I'm working on this now.

etwest:
> This code relies on sync to run without error. Before we can merge this we need to implement sync. I'm working on this now.

To an extent yes but in this setup in order for two workers to update the same node there will have to be two flushes of the same leaf node. This means that from the time the data enters the queue until the graph worker is done with processing it we will have to do at least 2M * tree height updates to the buffer tree triggering tree_height flushes. This assumes no other updates to any other nodes occur in that window. All this happening before the graph_worker is done is practically impossible.

However, if you finish the sync version and it works nicely I'm happy to have it included in this pull request. If it proves to be a pain then we don't need to include it.

EDIT: my math is wrong. if there are only updates to a single node than one flush will cause the buffer-tree to flush all the way to the leaf node after the root_node fills up once. We're still not seeing this happening in tests but on an adjacency node graph it could definitely happen. Thanks for working on the sync version. We need it and have needed it for awhile.

Victor-C-Zhang (approved):

PR #34, opened 2021-05-31
implicit syncing of sketch::operator+=

Victor-C-Zhang:
As advertised on the tin.

Adds an extra 40 bytes of overhead to a `Sketch` object. We should discuss if this is amenable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,7 @@ ExternalProject_Add(
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "circular_queue"
 
   BUILD_ALWAYS    OFF
   INSTALL_DIR     ${CMAKE_CURRENT_BINARY_DIR}/BufferTree/prefix
--- a/supernode.cpp
+++ b/supernode.cpp
@@ -68,7 +68,7 @@ void Supernode::batch_update(const std::vector<vec_t>& updates) {
    * this was slow (at least on small graph inputs).
    */
   std::vector<Sketch*> deltas(logn);
-  #pragma omp parallel for num_threads(GraphWorker::get_group_size())
+  #pragma omp parallel for num_threads(GraphWorker::get_group_size()) default(none) shared(deltas, updates)
   for (unsigned i = 0; i < sketches.size(); ++i) {
     deltas[i] = new Sketch(*sketches[i]);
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ ExternalProject_Add(
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "circular_queue"
```
Victor-C-Zhang, line 183: to get it to compile. we should change this back when merging to main

supernode.cpp
```diff
@@ -67,8 +67,18 @@ void Supernode::batch_update(const std::vector<vec_t>& updates) {
    * Considered using spin-threads and parallelism within sketch::update, but
    * this was slow (at least on small graph inputs).
    */
-  #pragma omp parallel for num_threads(GraphWorker::get_group_size())
+  std::vector<Sketch*> deltas(logn);
+  #pragma omp parallel for num_threads(GraphWorker::get_group_size()) default(none) shared(deltas, updates)
   for (unsigned i = 0; i < sketches.size(); ++i) {
-    sketches[i]->batch_update(updates);
+    deltas[i] = new Sketch(*sketches[i]);
```
etwest, line 73: Won't this cause problems? Say we start with a sketch which has updates for nodes = a,b,c in it. Then we apply a update which deletes node b. So our delta is now = a,c. When we below add this delta to the node sketch we will end up with a final answer of nodes = b which is wrong. The answer should be nodes = a,c

etwest: Additionally this is worrisome because we might make these copies in the middle of another node applying the deltas. So we would end up with some wonky results.

for this reason and the above concern we should start with null sketches rather than copies.

etwest: Marking necessary for merge

etwest (changes requested):

supernode.cpp
Victor-C-Zhang: The copy-constructor creates a blank sketch with the same meta-fields as the original. That is, `bucket_a` and `bucket_c` are zeroed-out, but the rest is the same.

Victor-C-Zhang: Although now that you mention it, is there a case where we want an actual copy-constructor?

etwest: Gotcha

etwest: Maybe at some point down the road but for now copying sketches is bad.

etwest (approved):
Looks good

PR #35, opened 2021-06-02
Code for multiple Connected Component Queries in one run

etwest:


CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ ExternalProject_Add(
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "multiple_CC"
```
etwest: Will need to change this before completing the pull request

graph.cpp
```diff
@@ -65,7 +66,7 @@ void Graph::batch_update(uint64_t src, const std::vector<uint64_t>& edges) {
 
 vector<set<Node>> Graph::connected_components() {
   bf->force_flush(); // flush everything in buffertree to make final updates
-  GraphWorker::stopWorkers(); // tell the workers to stop and wait for them to finish
+  GraphWorker::pause_workers(); // wait for the workers to finish applying the updates
```
etwest, line 69: Below this code there is a variable we use to disallow all future insertions to the graph after running the connected components. Will this be a problem or does code in some other branch address this?

graph.cpp
```diff
@@ -116,6 +117,14 @@ vector<set<Node>> Graph::connected_components() {
 #ifdef VERIFY_SAMPLES_F
   verifier.verify_soln(retval);
 #endif
+
+  // the GraphWorkers may now process more updates
+  // However! because sketches are modified in place continuing
+  // without being careful will cause bad sketch behavior!
+  // So, only give more updates after CC if you have restored
+  // the sketch state from before CC
+  GraphWorker::unpause_workers();
```
etwest: Another idea would be to have this unpause in its own graph function or to have some parameter to CC control whether to unpause. For now this seems like a reasonable approach. Any preferences or ideas?

graph.cpp
Victor-C-Zhang: Yes, since we want to allow multiple queries, this code and the associated exceptions/variables are unnecessary.

graph_worker.cpp
```diff
@@ -19,37 +29,74 @@ void GraphWorker::startWorkers(Graph *_graph, BufferTree *_bf) {
...
-		// and will enforce locking.
-		bool valid = bf->get_data(data);
-
-		if (valid)
-			graph->batch_update(data.first, data.second);
-		else if(shutdown)
+		if(shutdown)
 			return;
+		pause_lock.lock();
+		thr_paused = true; // this thread is currently paused
```
Victor-C-Zhang, line 80: I'm concerned this ordering of operations might cause us some problems when it comes to querying. if we push a set amount of data into the buffer tree and flush it, won't some tasks get stuck in the queue when we pause? for instance if the circular queue isn't empty.

graph_worker.cpp
```diff
@@ -19,37 +29,74 @@ void GraphWorker::startWorkers(Graph *_graph, BufferTree *_bf) {
...
-		// call get_data which will handle waiting on the queue
-		// and will enforce locking.
-		bool valid = bf->get_data(data);
-
-		if (valid)
-			graph->batch_update(data.first, data.second);
-		else if(shutdown)
+		if(shutdown)
 			return;
+		pause_lock.lock();
```
Victor-C-Zhang: an atomic variable is best suited for this semantic

graph_worker.cpp
```diff
@@ -19,37 +29,74 @@ void GraphWorker::startWorkers(Graph *_graph, BufferTree *_bf) {
...
+		while(!thr_paused) {
+			// call get_data which will handle waiting on the queue
+			// and will enforce locking.
+			bool valid = bf->get_data(data);
+
+			if (valid)
+				graph->batch_update(data.first, data.second);
+			else if(shutdown)
+				return;
+			else if(paused)
```
Victor-C-Zhang, line 97: are there race conditions or weird interactions if we don't synchronize reads of these variables?

graph.cpp
etwest: Now that I've thought about it. It could be a good idea to create an unpause function which either does the deserialization itself or needs to be passed the result of it which also unpauses the graph workers and sets this variable to false. This would enforce that updates cannot happen before the sketches are restored.

Thoughts?

graph_worker.cpp
etwest: Yes good point. I'll make the change

graph_worker.cpp
etwest: So this outer loop only runs when the thread pauses. The thread only pauses after failing to get data from the circular queue and if paused==true.

This means that so long as pause_workers is only called after force_flush that we should process all of the updates in the tree.

graph_worker.cpp
etwest: Yes but I don't believe it will cause problems.

Imagine we set paused to true as a worker is going through this if statement. It might be the case that the worker does not read paused==true so repeats. However, we will immediately exit from the call to bf->get_data (as the buffer tree operations are non-blocking once pause_workers is called) and check the if statement again.

graph_worker.cpp
Victor-C-Zhang: ok, and this is ok because we can be reasonably confident that force_flush will return after the inner loop has already begun?

graph_worker.cpp
Victor-C-Zhang: ok

graph.cpp
Victor-C-Zhang: i think this is ok for now. if we're going to pause and restart stream ingestion, we'll probably want to call this as part of a graph function like we do with everything else

graph.cpp
Victor-C-Zhang: I'm not sure what purpose this unpause function will serve. can you elaborate?

etwest: So basically, our normal code operates with the assumption that CC will only be called once. This new code adds the option to call it again. However, if you insert more updates after CC without resetting the sketches first things will get really weird. 

So this function I'm thinking of would be a graph function which enforces that you must reset the sketches before the graph code will accept another update.

etwest: It's the function I also discussed in the below comment

graph_worker.cpp
etwest: The inner loop should begin before force_flush is even called and if there's anything in the buffer tree when it is the inner loop must be executing in order to process that data.

Victor-C-Zhang: but in the case that we only have a few updates to process, we’ll rely on the force_flush timing right?

graph.cpp
Victor-C-Zhang: I see. Let’s do that

graph_worker.cpp
etwest: No because get_data will be a blocking call until we pause the GraphWorkers

graph.cpp
etwest: Okay there's one question I'd like to get yours and Kenny's take on. Should we setup this function to accept the original sketches as a parameter, to do the deserialization itself, or to trust that the user has deserialized before calling it?

etwest: Oh ... I guess we do have a use for a copy constructor... lol

kennyzzhang: I'm thinking that we should trust the user, since most of the time we'll just want to run the algorithm once and not care about saving the sketches

etwest: Okay, I wrote up that function. It's in the most recent pull request

graph_worker.cpp
Victor-C-Zhang: I’m a little confused. What’s the flow gonna be when, for instance, we push 10 updates into the buffer tree and then call `connected_components()`?

Victor-C-Zhang: Actually never mind i think i see it:
- we’ll call `force_flush`, and that puts some things into the circular queue
- the graph workers are notified and start pulling things out of the queue
- `force_flush` returns to the main thread
- we call pause, which blocks until all the graph workers are done

etwest: Yes exactly

etwest:
Ready to integrate

Victor-C-Zhang (approved):
Just change the branch of the buffer tree we pull from to `main`

PR #36, opened 2021-06-03
In-memory buffering

Victor-C-Zhang:
Add data structure to mimic buffer tree interface in-memory
Add toggle for fast buffer tree as a compilation definition `USE_FBT_F`

### DIFF
--- a/graph.cpp
+++ b/graph.cpp
@@ -6,7 +6,7 @@
 #include "include/util.h"
 #include "include/graph_worker.h"
 
-Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
+Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes), wq(1<<10, num_nodes) {
 #ifdef VERIFY_SAMPLES_F
   cout << "Verifying samples..." << endl;
 #endif
@@ -21,10 +21,15 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
   }
   num_updates = 0; // REMOVE this later
 
+#ifdef USE_FBT_F
   std::string buffer_loc_prefix = configure_system(); // read the configuration file to configure the system
   // Create buffer tree and start the graphWorkers
   bf = new BufferTree(buffer_loc_prefix, (1<<20), 8, num_nodes, GraphWorker::get_num_groups(), true);
   GraphWorker::startWorkers(this, bf);
+#else
+  GraphWorker::set_config(1,1); // TODO: actually get config
+  GraphWorker::startWorkers(this);
+#endif
 }
 
 Graph::~Graph() {
@@ -33,7 +38,9 @@ Graph::~Graph() {
   delete[] supernodes;
   delete[] parent;
   delete representatives;
+#ifdef USE_FBT_F
   delete bf;
+#endif
 }
 
 void Graph::update(GraphUpdate upd) {
@@ -45,8 +52,9 @@ void Graph::update(GraphUpdate upd) {
   std::swap(edge.first, edge.second);
   bf->insert(edge);
 #else
-  // TODO
+  wq.insert(edge);
   std::swap(edge.first, edge.second);
+  wq.insert(edge);
 #endif
 }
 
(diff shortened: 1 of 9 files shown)


graph.cpp
```diff
@@ -21,10 +21,16 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
   }
   num_updates = 0; // REMOVE this later
 
+#ifdef USE_FBT_F
   std::string buffer_loc_prefix = configure_system(); // read the configuration file to configure the system
   // Create buffer tree and start the graphWorkers
   bf = new BufferTree(buffer_loc_prefix, (1<<20), 8, num_nodes, GraphWorker::get_num_groups(), true);
   GraphWorker::start_workers(this, bf);
+#else
+  GraphWorker::set_config(1,1); // TODO: actually get config
```
etwest: Instead of doing this you should call configure_system

Just ignore the string it returns

etwest:
Before I approve this I'd like to see the tests for in memory buffering added to the GitHub actions

work_queue.cpp
```diff
@@ -0,0 +1,89 @@
+//
+// Created by victor on 6/2/21.
+//
+
+#include "include/work_queue.h"
+
+const unsigned first_idx = 2;
+
+WorkQueue::WorkQueue(uint32_t buffer_size, Node nodes) : buffer_size
+(buffer_size), cq(20,buffer_size*sizeof(Node)), buffers(nodes) {
```
etwest: Don't hard code the number of slots in the circular queue. This could cause us problems if we run with too many threads. The number of slots should be 2*num_groups

work_queue.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  // assume the first key is correct so extract it
+  Node key = serial_data[1];
+  data.first = key;
+
+  data.second.clear(); // remove any old data from the vector
+  uint32_t vec_len  = len / sizeof(Node);
+  data.second.reserve(vec_len); // reserve space for our updates
+
+  for (uint32_t j = first_idx; j < len / sizeof(Node); ++j) {
```
etwest: Probably doesn't make too much of a difference but you can replace len/sizeof(Node) with vec_len

work_queue.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void WorkQueue::flush(Node *buffer, uint32_t num_bytes) {
+  cq.push(reinterpret_cast<char *>(buffer), num_bytes);
+}
+
+insert_ret_t WorkQueue::insert(update_t upd) {
+  Node& idx = buffers[upd.first][0];
+  buffers[upd.first][idx] = upd.second;
+  ++idx;
```
etwest, line 34: This code assumes that it knows how big a Node is. The buffers array is of type unsigned long * which means this code only works if type(Node) = unsigned long.

This might not be a big problem but it could cause us problems if we ever want to change the type of Node

graph.cpp
Victor-C-Zhang: Sure, I guess we can just pass a nonsense string to the prefix

work_queue.cpp
Victor-C-Zhang: I don't think so. The buffers array is of type `Node*`, and it will change if we change the type of `Node`.

Victor-C-Zhang:
> Before I approve this I'd like to see the tests for in memory buffering added to the GitHub actions

Added a CI step to run the testing suite with in-memory buffering.

graph.cpp
etwest: Yeah, we can just leave the path_prefix as the default or even remove it entirely from the configuration file. The system will throw a warning but it won't make an actual difference

work_queue.cpp
etwest: Okay, can we change line 12 to also be a static cast to Node* then?

graph.cpp
```diff
@@ -20,11 +20,15 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
...
   }
   num_updates = 0; // REMOVE this later
-
   std::string buffer_loc_prefix = configure_system(); // read the configuration file to configure the system
+#ifdef USE_FBT_F
   // Create buffer tree and start the graphWorkers
   bf = new BufferTree(buffer_loc_prefix, (1<<20), 8, num_nodes, GraphWorker::get_num_groups(), true);
   GraphWorker::start_workers(this, bf);
+#else
+  wq = new WorkQueue(1 << 10, num_nodes, 2*GraphWorker::get_num_groups());
```
etwest: Is 1 << 10 the size of the buffer?
If so, that's too small. The size of each node sketch is much larger than a kilobyte once the number of nodes is greater than 11. We want the buffer to be the size of a node sketch, or 24 * log_2^3(N) rounded to the nearest Node.
Instead of a static approach it's probably worth while to calculate the value we want.

work_queue.cpp
Victor-C-Zhang: oops

etwest (approved):
Looks good to me

PR #39, opened 2021-06-03
Half Repetitions

etwest:


etwest:
Kenny - The change I made here is correct for half repetitions right? 

If so, why are tests failing and how can we address that issue?

kennyzzhang:
Ah, the sampling test is still comparing against a fixed probability. A quick fix is to go into sketch_test.cpp, and for SketchTestSample and SketchTestAddition, explicitly set bucket repetitions to 1.

etwest:
Okay so these failing tests aren't an indication of something wrong right? If so can we get this pull request approved and merged?

Victor-C-Zhang:
I'd be a little more at ease if we set the probability based on repetitions, but in the interest of time I think this is fine. I've opened an issue for this that should be resolved once we're out of crunch mode.

Victor-C-Zhang (approved):
See comment and related issue

PR #41, opened 2021-06-05
Chore: bugfixes

Victor-C-Zhang:
- Correct size of nodes in in-memory buffering system (thanks @etwest for pointing this out)
- Use correct size types for vec_t and vec_hash_t (64, 32 respectively)
- Add OpenMP parallelization for `SupernodeTestSuite.TestConcurrency`

### DIFF
--- a/graph.cpp
+++ b/graph.cpp
@@ -26,9 +26,11 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
   bf = new BufferTree(buffer_loc_prefix, (1<<20), 8, num_nodes, GraphWorker::get_num_groups(), true);
   GraphWorker::start_workers(this, bf);
 #else
+  unsigned num_buckets = bucket_gen(num_nodes*num_nodes, 1); // TODO: don't fix num_bucket_factor
+  unsigned num_guesses = guess_gen(num_nodes*num_nodes);
   unsigned node_size = sizeof(Supernode) + double_to_ull(log2(num_nodes))*(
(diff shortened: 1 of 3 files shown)


graph.cpp
```diff
@@ -26,9 +26,11 @@ Graph::Graph(uint64_t num_nodes): num_nodes(num_nodes) {
   bf = new BufferTree(buffer_loc_prefix, (1<<20), 8, num_nodes, GraphWorker::get_num_groups(), true);
   GraphWorker::start_workers(this, bf);
 #else
+  unsigned num_buckets = bucket_gen(num_nodes*num_nodes, 1); // TODO: don't fix num_bucket_factor
+  unsigned num_guesses = guess_gen(num_nodes*num_nodes);
   unsigned node_size = sizeof(Supernode) + double_to_ull(log2(num_nodes))*(
-        sizeof(Sketch) + bucket_gen(num_nodes*num_nodes, 1)*sizeof(vec_t) +
-        guess_gen(num_nodes)*sizeof(vec_hash_t));
+        sizeof(Sketch)
+        + num_buckets*num_guesses*(sizeof(vec_t) + sizeof(vec_hash_t)));
```
etwest, line 33: Have you verified that this is reasonably close to the size we expect?

etwest (approved):
If you answer the question above then this looks good to me.

graph.cpp
Victor-C-Zhang: Yes, testing with `n` up to 2^15, this size is between 3 times and 2 times as large as the naive `24*log_2^3(n)` estimate.

PR #43, opened 2021-06-07
buffer tree having 16 children per node is more efficient

etwest:
Just a small configuration change.


Victor-C-Zhang:
how did you determine this?

Victor-C-Zhang (approved):

etwest:
> how did you determine this?

I ran a variety of experiments on the buffer tree end of things and B=16 was better than b=8

PR #46, opened 2021-07-19
Parallelize post-processing

Victor-C-Zhang:
Write `parallel_connected_components` function using OpenMP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,6 +266,7 @@ endif()
 add_executable(experiment
   test/test_runner.cpp
   test/experiment/BCHMK_Graph.cpp
+  test/experiment/BCHMK_post_proc.cpp
   test/util/graph_verifier.cpp
   test/util/graph_gen.cpp
   graph.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON)
 
 #Uncomment to enable debug
-#set(ENV{DEBUG} ON)
+set(ENV{DEBUG} ON)
 
 project(test)
 
@@ -169,7 +169,6 @@ link_directories(${CMAKE_CURRENT_BINARY_DIR}/BufferTree/prefix/lib)
 # AVAILABLE COMPILATION DEFINITIONS:
 # VERIFY_SAMPLES_F   Use a deterministic connected-components algorithm to
 #                    verify post-processing.
-#                    INCOMPATIBLE with PARALLEL_PP_F.
 #
 # USE_NATIVE_F       ???
 #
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON)
 
 #Uncomment to enable debug
-#set(ENV{DEBUG} ON)
+set(ENV{DEBUG} ON)
```
etwest: turn off :)

graph.cpp
```diff
@@ -141,6 +141,89 @@ vector<set<Node>> Graph::connected_components() {
...
+      if (size[a] < size[b]) std::swap(a,b);
+      to_remove.push_back(b);
+      parent[b] = a;
+      size[a] += size[b];
+      supernodes[a]->merge(*supernodes[b]);
+    }
+    if (!to_remove.empty()) modified = true;
+    sort(to_remove.begin(), to_remove.end());
+
+    // 2-pointer to find set difference
```
etwest, line 197: Can you explain what this set difference thing is doing?

graph.cpp
```diff
@@ -141,6 +141,89 @@ vector<set<Node>> Graph::connected_components() {
...
+  Node size[num_nodes];
+  vector<Node> reps(num_nodes);
+  fill(size, size + num_nodes, 1);
+  for (Node i = 0; i < num_nodes; ++i) {
+    reps[i] = i;
+  }
+
+  do {
+    modified = false;
+    #pragma omp parallel for default(none) shared(query, reps)
```
etwest, line 169: It seems like we're only running a very small part of this in parallel. Is this the intensive part of the work?

Victor-C-Zhang: It’s the only part that can be efficiently parallel IMO. Since merging deltas requires updating the DSU arrays, they have to be done sequentially.

etwest: What are the DSU arrays?

Victor-C-Zhang: Disjoint set union; keeps track of the supernodes.

graph.cpp
Victor-C-Zhang: Essentially, we're treating `reps` and `to_remove` as sets, and we're finding `temp_diff = reps \ to_remove`

graph.cpp
etwest: So I looked a bit into the theory on this. It seems there is a wait-free version of these disjoint set union structures. However, I don't think we have time before the deadline to make this change.

graph.cpp
etwest: Gotcha

graph.cpp
etwest: So maybe we can open an issue about this and address it after the deadline.

etwest (approved):
Looks good as far as I can tell. With a lot more work we could get more performance down the road. For now this is good though.

PR #48, opened 2021-07-26
Different seeds for sketches inside supernode

kennyzzhang:


Victor-C-Zhang:
Take a look at the failing tests
```
[ RUN      ] SupernodeTestSuite.TestSampleDeleteGrinder
unknown file: Failure
C++ exception with description "This sketch has already been sampled!" thrown in the test body.
```

kennyzzhang:
Found the bug.

When a `Sketch` finds no good buckets, it throws `NoGoodBucketException`. Due to how `Supernode::sample()` is written, this `NoGoodBucketException` prevents `Supernode` from correctly incrementing its index. Normally this isn't an issue, because normally `NoGoodBucketException` means the algorithm fails and stops, and it never queries the `Supernode` again. However, `TestSampleInsertGrinder` and `TestSampleDeleteGrinder` catch and ignore `NoGoodBucketException`, so when these tests blindly continue to query the `Supernode`, they resample the `Sketch` at the old unincremented index.

supernode.cpp
```diff
@@ -10,7 +10,7 @@ Supernode::Supernode(uint64_t n, long seed): sketches(log2(n)), idx(0), logn(log
 (n)), seed(seed), n(n) {
   // generate logn sketches for each supernode (read: node)
   for (int i=0;i<logn;++i)
-    sketches[i] = new Sketch(n*n, seed);
+    sketches[i] = new Sketch(n*n, seed++);
```
Victor-C-Zhang, line 13: which `seed` is this incrementing? if it's the Supernode object's `seed`, we're going to have a bad time

kennyzzhang: Because of variable shadowing, it's not the `Supernode` object's `seed`.

Victor-C-Zhang (approved):
This is smelly, but it’ll work

PR #51, opened 2021-08-27
test: use half repetitions for sketch unit tests

Victor-C-Zhang:
Use sketch default repetitions to detect big problems with sketching if they happen


etwest:
So, this didn't pass our unit tests. So we probably don't want to make this change. Likely, things are failing because our graphs we're testing things on are really small. Maybe we could come up with one test for half repetitions which uses a larger input?

etwest:
Closing because sketch architecture has changed such that half-repetitions are no longer a part of the system.

PR #55, opened 2021-09-09
Memory Architecture Changes to improve Sketch performance

etwest:


etwest:
There are conflicts in this branch that I will fix later. We need to be careful to preserve the first column optimization while also keeping the memory architecture changes.

Victor-C-Zhang:
> There are conflicts in this branch that I will fix later. We need to be careful to preserve the first column optimization while also keeping the memory architecture changes.

Will look at your changes except for sketch

CMakeLists.txt
```diff
@@ -271,12 +273,13 @@ add_executable(experiment
...
   graph.cpp
   supernode.cpp
+  work_queue.cpp
   graph_worker.cpp
   l0_sampling/sketch.cpp
   l0_sampling/update.cpp
   util.cpp)
 target_link_libraries(experiment PRIVATE xxHash::xxhash GTest::gtest)
-target_compile_definitions(experiment PRIVATE USE_FBT_F)
+# target_compile_definitions(experiment PRIVATE USE_FBT_F)
```
Victor-C-Zhang, line 282: so are our experiments not gonna use FBT?

graph.cpp
```diff
@@ -65,7 +99,7 @@ void Graph::update(GraphUpdate upd) {
 }
 
 Supernode* Graph::generate_delta_node(uint64_t node_n, long node_seed, uint64_t
-                                src, const std::vector<uint64_t>& edges) {
+							 src, const std::vector<uint64_t>& edges, Supernode *delta_loc) {
```
Victor-C-Zhang, line 103: if we're using placement construction, why are we returning anything?

graph.cpp
```diff
@@ -77,15 +111,22 @@ Supernode* Graph::generate_delta_node(uint64_t node_n, long node_seed, uint64_t
...
   }
-  return Supernode::delta_supernode(node_n, node_seed, updates);
+  return Supernode::delta_supernode(node_n, node_seed, updates, delta_loc);
 }
-void Graph::batch_update(uint64_t src, const std::vector<uint64_t>& edges) {
+void Graph::batch_update(uint64_t src, const std::vector<uint64_t>& edges, Supernode *delta_loc) {
   if (update_locked) throw UpdateLockedException();
+  // page in the supernodes
+  Supernode::page_in(delta_loc);
+  Supernode::page_in(supernodes[src]);
```
Victor-C-Zhang: unneeded?

graph.cpp
```diff
@@ -77,15 +111,22 @@ Supernode* Graph::generate_delta_node(uint64_t node_n, long node_seed, uint64_t
...
-  auto* delta_node = generate_delta_node(supernodes[src]->n,
-                                         supernodes[src]->seed, src, edges);
+  Supernode* delta_node = generate_delta_node(supernodes[src]->n,
+					supernodes[src]->seed, src, edges, delta_loc);
   supernodes[src]->apply_delta_update(delta_node);
-  delete(delta_node);
+
+  // page out the supernodes
+  Supernode::page_out(delta_loc);
+  Supernode::page_out(supernodes[src]);
```
Victor-C-Zhang: Similarly here

graph.cpp
```diff
@@ -104,6 +145,11 @@ vector<set<Node>> Graph::connected_components() {
   GraphVerifier verifier {cum_in};
 #endif
 
+  // force DSU structure into memory
+  // if this call fails, there is no problem since unlock won't throw any
+  // errors if we try to unlock un-locked memory
+  mlock(parent, num_nodes*sizeof(Node));
```
Victor-C-Zhang: same

graph.cpp
```diff
@@ -136,6 +182,10 @@ vector<set<Node>> Graph::connected_components() {
     if (!removed.empty()) modified = true;
     for (Node i : removed) representatives->erase(i);
   } while (modified);
+
+  // release DSU lock
+  munlock(parent, num_nodes*sizeof(Node));
```
Victor-C-Zhang: same

include/graph.h
```diff
@@ -95,7 +101,13 @@ class Graph{
   std::atomic<uint64_t> num_updates;
 
   static Supernode* generate_delta_node(uint64_t node_n, long node_seed, uint64_t src,
```
Victor-C-Zhang: document please

Victor-C-Zhang: also, why are we both taking a `delta_loc` and returning a pointer?

include/graph_worker.h
```diff
@@ -39,9 +40,9 @@ class GraphWorker {
 
 	// manage threads
 #ifdef USE_FBT_F
-  static void start_workers(Graph *_graph, BufferTree *_db); // start the graph workers
+  static void start_workers(Graph *_graph, BufferTree *_db, long _supernode_size); // start the graph workers
```
Victor-C-Zhang: document please

include/graph_worker.h
```diff
@@ -39,9 +40,9 @@ class GraphWorker {
...
 #else
-  static void start_workers(Graph *_graph, WorkQueue *_wq); // start the graph workers
+  static void start_workers(Graph *_graph, WorkQueue *_wq, long _supernode_size); // start the graph workers
```
Victor-C-Zhang: same here

include/sketch.h
```diff
@@ -18,35 +20,65 @@
...
+  Sketch(long seed);
+  Sketch(long seed, std::fstream &binary_in);
+
+public:
+  static SketchUniquePtr makeSketch(const Sketch &old);
+  static SketchUniquePtr makeSketch(long seed);
+  static SketchUniquePtr makeSketch(long seed, std::fstream &binary_in);
+  static Sketch* makeSketch(void* loc, long seed);
+  static Sketch* makeSketch(void* loc, long seed, std::fstream &binary_in);
+  static Sketch* makeSketch(void* loc, long seed, double num_bucket_factor, std::fstream &binary_in);
```
Victor-C-Zhang: what's the point of unique pointers?

include/sketch.h
```diff
@@ -18,35 +20,65 @@
...
+    num_buckets = bucket_gen(n, num_bucket_factor);
+    num_guesses = guess_gen(n);
+    num_elems = num_buckets * num_guesses;
+    return num_elems;
+  }
+  
+  inline static size_t sketchSizeof()
+  { return sizeof(Sketch) + get_num_elems() * (sizeof(vec_t) + sizeof(vec_hash_t)) - sizeof(char); }
+  
+  inline double get_bucket_factor()
```
Victor-C-Zhang: this field is public, so this getter is redundant

include/sketch.h
```diff
@@ -18,35 +20,65 @@
...
+
+public:
+  static SketchUniquePtr makeSketch(const Sketch &old);
+  static SketchUniquePtr makeSketch(long seed);
+  static SketchUniquePtr makeSketch(long seed, std::fstream &binary_in);
+  static Sketch* makeSketch(void* loc, long seed);
+  static Sketch* makeSketch(void* loc, long seed, std::fstream &binary_in);
+  static Sketch* makeSketch(void* loc, long seed, double num_bucket_factor, std::fstream &binary_in);
 
+  inline static unsigned get_num_elems() {
```
Victor-C-Zhang: num_elems is public, so either delete this function or make it private

include/supernode.h
```diff
@@ -1,38 +1,92 @@
...
+
 #include "sketch.h"
 
 using namespace std;
 
 typedef uint64_t Node;
 typedef std::pair<Node, Node> Edge;
 
+// #define USE_MLOCK
+
```
Victor-C-Zhang: get rid of

include/supernode.h
```diff
@@ -1,38 +1,92 @@
...
 /**
  * This interface implements the "supernode" so Boruvka can use it as a black
  * box without needing to worry about implementing l_0.
  */
 class Supernode {
-  /* collection of logn sketches to query from, since we can't query from one
-     sketch more than once */
-  vector<Sketch*> sketches;
+  static uint32_t bytes_size;
```
Victor-C-Zhang: document this var

include/supernode.h
```diff
@@ -1,38 +1,92 @@
...
+   * @param seed      the (fixed) seed value passed to each supernode.
+   * @param binary_in A stream to read the file from
+   */
+  Supernode(uint64_t n, long seed, std::fstream &binary_in);
+
+public:
+  using SupernodeUniquePtr = std::unique_ptr<Supernode, std::function<void(Supernode*)>>;
+  static SupernodeUniquePtr makeSupernode(uint64_t n, long seed);
+  static SupernodeUniquePtr makeSupernode(uint64_t n, long seed, std::fstream &binary_in);
+  static Supernode* makeSupernode(void* loc, uint64_t n, long seed);
```
Victor-C-Zhang: why do we need so many constructors? surely we can just use one for normal construction and maybe another for reheated construction

include/supernode.h
```diff
@@ -1,38 +1,92 @@
...
+
   ~Supernode();
 
+  inline Sketch* get_sketch(size_t i)
+  { return reinterpret_cast<Sketch*>(sketch_buffer + i * sketch_size); }
+
+  inline const Sketch* get_sketch(size_t i) const
```
Victor-C-Zhang, line 65: why 2 copies of this function?

include/supernode.h
```diff
@@ -1,38 +1,92 @@
...
+    mlock(sn, bytes_size);
+#endif
+  }
+
+#ifdef USE_MLOCK 
+  static inline void page_out(Supernode* sn) 
+  { munlock(sn, bytes_size); }
+#else 
+  static inline void page_out(Supernode *) {}
+#endif
```
Victor-C-Zhang: remove

include/supernode.h
```diff
@@ -71,8 +125,21 @@ class Supernode {
    * @param updates the batch of updates to apply.
    * @return
    */
-  static Supernode* delta_supernode(uint64_t n, long seed, const
+  static Supernode::SupernodeUniquePtr delta_supernode(uint64_t n, long seed, const
```
Victor-C-Zhang: This is (or at least should be) useless. Our graph workers have pre-allocated buffers for delta nodes. There should be no reason to need to allocate space on the fly like this.

l0_sampling/sketch.cpp
```diff
@@ -1,29 +1,74 @@
...
+Sketch::SketchUniquePtr Sketch::makeSketch(const Sketch &old) {
+  return makeSketch(old.seed);
+}
+
+
+/*
+ * Static functions for creating sketches without a provided memory location
+ * these are useful for unit tests
+ */
+Sketch::SketchUniquePtr Sketch::makeSketch(long seed) {
```
Victor-C-Zhang: If these are only used for unit tests, they should be in the testing/utility files, not here.

l0_sampling/sketch.cpp
```diff
@@ -1,29 +1,74 @@
...
+Sketch::SketchUniquePtr Sketch::makeSketch(long seed, std::fstream &binary_in) {
+  void* loc = malloc(sketchSizeof());
+  return SketchUniquePtr(makeSketch(loc, seed, binary_in), [](Sketch* s){ free(s); });
+}
+
+/*
+ * Static functions for creating sketches with a provided memory location.
+ * We use these in the production system to keep supernodes virtually contiguous.
+ */
+Sketch* Sketch::makeSketch(void* loc, long seed) {
```
Victor-C-Zhang, line 15: again, why are we returning the pointer to a location we provide?

l0_sampling/sketch.cpp
```diff
@@ -75,19 +116,26 @@ Sketch &operator+= (Sketch &sketch1, const Sketch &sketch2) {
...
-  return sketch1.n == sketch2.n && sketch1.seed == sketch2.seed &&
-    sketch1.num_bucket_factor == sketch2.num_bucket_factor &&
-    sketch1.bucket_a == sketch2.bucket_a &&
-    sketch1.bucket_c == sketch2.bucket_c &&
-    sketch1.already_quered == sketch2.already_quered;
+  if (sketch1.seed != sketch2.seed || sketch1.already_quered != sketch2.already_quered) 
+    return false;
+
+  for (size_t i = 0; i < Sketch::num_elems; ++i)
+  {
```
Victor-C-Zhang: nit(code style): open brace on prev line

supernode.cpp
```diff
@@ -1,28 +1,60 @@
...
+  SupernodeUniquePtr ret(new (loc) Supernode(n, seed, binary_in), [](Supernode* s){ free(s); });
+  return ret;
+}
+
+Supernode* Supernode::makeSupernode(void* loc, uint64_t n, long seed) {
+  return new (loc) Supernode(n, seed);
 }
 
 Supernode::~Supernode() {
-  for (int i=0;i<logn;++i)
```
Victor-C-Zhang, line 17: can you remind me why we don't need to worry about de-allocating?

supernode.cpp
```diff
@@ -1,28 +1,60 @@
 #include <stdexcept>
```
Victor-C-Zhang, line 1: comments in this file follow from comments in the header files

test/graph_test.cpp
```diff
@@ -88,3 +88,49 @@ TEST(GraphTestSuite, TestCorrectnessOnSmallSparseGraphs) {
...
+    // for (int i = 0; i < log2(n); i++) {
+    //   ASSERT_EQ(g.supernodes[i]->logn, reheated.supernodes[i]->logn);
+    //   //ASSERT_EQ(g.supernodes[i]->idx, reheated.supernodes[i]->idx);
+    //   for (int j = g.supernodes[i]->idx;j < g.supernodes[i]->logn;++j) {
+    //     if(!(*g.supernodes[i]->get_sketch(j) == *reheated.supernodes[i]->get_sketch(j))) {
+    //       printf("sketches at index %i do not match of supernode %i\n", j, i);
+    //       ASSERT_EQ(1, 0);
+    //     }
+    //   }
+    // }
```
Victor-C-Zhang: why is this commented out?

test/supernode_test.cpp
```diff
@@ -167,73 +166,55 @@ TEST_F(SupernodeTestSuite, TestBatchUpdate) {
...
 
 void apply_delta_to_node(Supernode* node, const std::vector<vec_t>& updates) {
-  auto* delta_node = Supernode::delta_supernode(node->n, node->seed, updates);
-  node->apply_delta_update(delta_node);
-  delete(delta_node);
+  auto delta_node = Supernode::delta_supernode(node->n, node->seed, updates);
+  node->apply_delta_update(delta_node.get());
 }
 
-TEST_F(SupernodeTestSuite, TestConcurrency) {
```
Victor-C-Zhang, line 196: why did we get rid of this test?

test/util/efficient_gen/edge_gen.cpp
```diff
@@ -2,7 +2,11 @@
 #include <algorithm>
 #include <boost/multiprecision/cpp_int.hpp>
 #include "efficient_gen.h"
-#include "../../../include/graph.h"
+
+enum UpdateType {
+  INSERT = 0,
+  DELETE = 1,
+};
```
Victor-C-Zhang: either get rid of this enum or the one in `graph.h`.

Victor-C-Zhang:
There are some design choices I'm not sold on, specifically our use of smart pointers in addition to normal pointers and constructors. Maybe this is something we'll need to sync on.

CMakeLists.txt
etwest: The branch where I actually have the experiments uses two executables one with FBT and one without.

graph.cpp
etwest: I'm not sure yet. I'll run a quick experiment and find out

graph.cpp
etwest: I suppose we don't need to. Do you think we should make this a void function?

graph.cpp
etwest: Oops thought I removed this. I'll do that now.

include/graph.h
etwest: Will do. And I'll make it a void function

graph.cpp
etwest: I'll just make it void

l0_sampling/sketch.cpp
etwest: Here I actually think this is clearer. Just because these make functions are acting as constructors so returning a pointer to the 'new' object makes sense. Otherwise some of these return sketches and some don't which would be harder to understand.

l0_sampling/sketch.cpp
etwest: They are only used for unit tests

supernode.cpp
etwest: Because the memory is contiguous and is freed by the graph code at the top.

etwest: So the calls to delete are redundant

test/graph_test.cpp
etwest: Because it was a part of our debugging not a part of the actual test. I will remove the code.

include/supernode.h
etwest: Maybe, depends if it helps with insertion performance. It probably won't but I want to check.

include/supernode.h
etwest: Tyler did this, I think its so that we can use a const * when we need one and a non-const otherwise

include/supernode.h
etwest: 1. Create new empty super node 
2. Create new super node from reheat
3. Initialize a memory location as a empty super node. (used in generate_delta_node)

include/sketch.h
etwest: We don't have to free them. Their free'd when they go out of scope.

include/sketch.h
etwest: Okay

include/sketch.h
etwest: Okay

include/supernode.h
etwest: Will do

include/supernode.h
etwest: This was left in for testing purposes.

l0_sampling/sketch.cpp
etwest: I'll change this

supernode.cpp
etwest: Also because they're probably unique pointers

test/supernode_test.cpp
etwest: I have no idea. I'll try to add it back in.

test/util/efficient_gen/edge_gen.cpp
etwest: The point of this enum was to remove the dependency on graph.h because that was causing the compilation to fail on my system.

include/sketch.h
Victor-C-Zhang: But we never use them in the non-testing codebase.

CMakeLists.txt
Victor-C-Zhang: so is this executable useless?

test/util/efficient_gen/edge_gen.cpp
Victor-C-Zhang: There are now two copies of this enum. There should only be one.

l0_sampling/sketch.cpp
Victor-C-Zhang: So they should go into a testing file as a helper function.

include/supernode.h
Victor-C-Zhang: Can you refactor this as a utility/helper function into our testing code then?

include/supernode.h
Victor-C-Zhang: Not too sold on why we're using const, but it'll be fine for now.

include/supernode.h
Victor-C-Zhang: Mixing smart pointers and normal pointers doesn't sit well with me. I think we should get rid of either (1) or (3)

supernode.cpp
```diff
@@ -1,28 +1,60 @@
...
+
+  // read logn sketches from file for each supernode (read: node)
+  for (int i = 0; i < logn; ++i) {
+    Sketch::makeSketch(get_sketch(i), seed++, binary_in);
+  }
+}
+
+Supernode::SupernodeUniquePtr Supernode::makeSupernode(uint64_t n, long seed) {
+  Sketch::n = n * n;
+  set_size(n);
```
Victor-C-Zhang: this should not happen in this constructor. this will be called `n - 1` times more than it should.

CMakeLists.txt
etwest: For this branch it's fine

include/binary_graph_stream.h
```diff
@@ -0,0 +1,58 @@
...
+
+    inline GraphUpdate get_edge() {
+        UpdateType u = (UpdateType) *buf;
+	buf++;
+
+	uint32_t a;
+	std::memcpy(&a, buf, sizeof(uint32_t));
+	buf += 4;
+	uint32_t b;
+	std::memcpy(&b, buf, sizeof(uint32_t));
```
Victor-C-Zhang: u can probably do `memcpy(buf+4)` and increment by 8

etwest: Will do

include/supernode.h
etwest: I'll convert everything to standard pointers

include/supernode.h
etwest: I just removed it.

l0_sampling/sketch.cpp
etwest: Did that

test/supernode_test.cpp
etwest: Added back in

supernode.cpp
etwest: I don't know if this will make much of a difference because as previously discussed we only use this call at the beginning of the program. So yes our initialization might take a tiny bit longer but our ingestion rate will be unaffected.

There is benefit to making the process of setting the size of a super node transparent to the caller.

Victor-C-Zhang: I would disagree. We should only have one (fixed) value of `n` and fixed values of `Sketch::n` and sketch size. Will it require more architectural changes to kick this out of supernode constructor?

etwest: I guess you're right. I think what we would want to do is have a static configure function for Sketch, Supernode, and Graph which sets these parameters appropriately. 

We can also set it up so calling the Graph configure function calls super node which calls Sketch

I'll code that up and push it.

Victor-C-Zhang (approved):
LGTM

PR #56, opened 2021-09-13
refactor: fix up codebase for review

### LINKED ISSUE #37: USE_NATIVE_F is a mess
Currently, we're using 32 bit types for `bucket_a` and `bucket_c` so in theory we should be able to support only 32-bit numbers of nodes. But we use type `Node = uint64_t` for everything else, which means we should instead be using 64 bits for `bucket_a`. So far we haven't noticed anything wrong because our tests are small, but this *must* be fixed before we can do any large scale testing.

Victor-C-Zhang:
- deleted 128-bit files
- deleted flamegraphs
- deleted `main.cpp`, `l0sampling.cpp`
- deleted `test/experiment/par.tex`
- move `l0_sampling` includes to separate folder to mirror source file structure
- move source files into `src` directory
- janitor all files
- remove`USE_NATIVE_F` and close #37 
- convert stray tab indents to spaces
- cum -> cumul


etwest:
We should talk about this at the meeting today with everyone. I think there are some other changes we would want to make.

include/bucket.h
```diff
@@ -61,7 +61,7 @@ namespace Bucket_Boruvka {
    * @param update_hash The hash of the update index, generated with Bucket::index_hash.
    */
   inline static void update(vec_t& a, vec_hash_t& c, const vec_t& update_idx, const vec_hash_t& update_hash);
-}
+} // namespace Bucket_Boruvka
```
etwest, line 64: Why is this here?

Victor-C-Zhang: to mark the end of the namespace declaration. boost does a similar thing in their repos

include/work_queue.h
```diff
@@ -1,13 +1,6 @@
...
-//
-
-#ifndef TEST_WORK_QUEUE_H
-#define TEST_WORK_QUEUE_H
-
+#pragma once
 #include <buffer_tree.h>
-#include "supernode.h"
-// TODO: switch references from Node to node_t in types, change CMake flags
+// TODO: switch references from Node to node_t in types
```
etwest, line 3: Can we get rid of this?

Victor-C-Zhang: we haven't done it though. there's still `Node` type in the buffer tree repo that we use

test/util/efficient_gen/efficient_gen.h
```diff
@@ -1,14 +1,9 @@
...
 // insert, delete based on a geometric distribution with ratio p
 // i.e. p% of edges will be deleted, p^2% will be re-inserted, p^3 will be re-deleted
 // until 1 element is left
 void insert_delete(double p, std::string in_file, std::string out_file);
 
-void write_cum(std::string stream_f, std::string cum_f);
-
-
-#endif //TEST_EFFICIENT_GEN_H
+void write_cum(const std::string& stream_f, const std::string& cum_f);
```
etwest: We might want to think about avoiding 'cum' in our repo... maybe

include/bucket.h
etwest: Ok

include/work_queue.h
etwest: Guess so

test/util/efficient_gen/efficient_gen.h
Victor-C-Zhang: ok

etwest (approved):
looks good

PR #59, opened 2021-10-07
Statistical testing

etwest:


CMakeLists.txt
```diff
@@ -251,6 +251,36 @@ else ()
...
+  test/util/graph_gen.cpp
+  src/graph.cpp
+  src/supernode.cpp
+  src/graph_worker.cpp
+  src/work_queue.cpp
+  src/l0_sampling/sketch.cpp
+  src/l0_sampling/update.cpp
+  src/util.cpp)
+target_link_libraries(statistical_test PRIVATE xxHash::xxhash GTest::gtest)
+target_compile_definitions(statistical_test PRIVATE VERIFY_SAMPLES_F)
```
Victor-C-Zhang: if this is running as our cron job, shouldn't we have both with and without gutter tree?

test/statistical_testing/graph_testing.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return failures;
+}
+
+int medium_with_iso_test(int runs) {
+    int failures = 0;
+    for (int i = 0; i < runs; i++) {
+        generate_stream({2048,0.002,0.5,0,"./sample.txt","./cumul_sample.txt"});
+        std::fstream graph_file{"./sample.txt"};
+        std::fstream cumul_file{"./cumul_sample.txt"};
+        graph_file.write("2070", 4); // increase the node size to create iso nodes
```
Victor-C-Zhang, line 52: what are iso nodes?

test/statistical_testing/analyze_results.py
```diff
@@ -0,0 +1,72 @@
+
+import numpy as np
+import argparse
+from scipy.stats import ttest_ind
+
+def check_error(test_name, test_result_file, expected_result_file, confidence=0.95):
```
Victor-C-Zhang, line 6: why do you take a confidence argument if you never use it?

test/statistical_testing/graph_testing.cpp
etwest: So this is basically a test which approximates our kron streams where there are a few nodes which are guaranteed to not have any neighbors.

test/statistical_testing/analyze_results.py
etwest: I want to use it but I don't know how to determine the z-value from the confidence value. Do you?

CMakeLists.txt
etwest: Yes I suppose so, I'll implement that change

test/statistical_testing/analyze_results.py
Victor-C-Zhang: Suppose our confidence is 2c. Then we want to find the value of x such that the Normal CDF takes value .5+c

The CDF is given
.5[1 + erf(x/sqrt2)]

You can find this x value by any means, for instance Newton or even a dumb binary search.
I can implement this tonight if you’d like.

etwest:
Thanks Victor

CMakeLists.txt
etwest: Done

Victor-C-Zhang (approved):
Approving changes, but still worried about test failure rates

PR #60, opened 2021-10-12
refactor: use external node type

Victor-C-Zhang:


CMakeLists.txt
```diff
@@ -146,13 +146,33 @@ if(NOT Boost_FOUND)
...
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/GraphZeppelinCommon/prefix/include)
+
 # Install BufferTree Project
 ExternalProject_Add(
   BufferTree
 
   PREFIX          BufferTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "commonComp"
```
etwest, line 175: Change this before merging

etwest (approved):
Approve pending change to branch

etwest (changes requested):
Actually, I'm a bit concerned about this. Currently we have two types in the codebase. There's a 64 bit value which we use for the bucket_a values which we call Node (or something) and there's a 32 bit value which we use for node_id_t and bucket_c. I think these changes would need to somehow incorporate both of these types.

Victor-C-Zhang:
> Actually, I'm a bit concerned about this. Currently we have two types in the codebase. There's a 64 bit value which we use for the bucket_a values which we call Node (or something) and there's a 32 bit value which we use for node_id_t and bucket_c. I think these changes would need to somehow incorporate both of these types.

I don’t think `node_id_t` is used in the streaming repo. iirc it’s used to number buffers in the buffer tree only (?)

etwest:
> > Actually, I'm a bit concerned about this. Currently we have two types in the codebase. There's a 64 bit value which we use for the bucket_a values which we call Node (or something) and there's a 32 bit value which we use for node_id_t and bucket_c. I think these changes would need to somehow incorporate both of these types.
> 
> I don’t think `node_id_t` is used in the streaming repo. iirc it’s used to number buffers in the buffer tree only (?)

It's also used by the standalone buffering and we might want to think about expanding its use. 64bit bucket_a only works if we have 32 bit node ids

PR #61, opened 2021-10-18
Continuous Correctness Testing

kennyzzhang:
There's still an issue with the supernode copying. From reading the current code, I was under the impression that I could memcpy the sketch_buffer to another place, then after running connected components, free the used supernodes and point the graph to the copy. But evidently somewhere somethings goes wrong, and it segfaults after the first correctness check.


etwest:
This is an issue we’ve dealt with before. Part of the supernode/sketch structure is a pointer to where the associated data lives. Your copy constructor will need to be a little more intelligent that a direct memcpy. I can write some code for you if you want.

kennyzzhang:
> 
> 
> This is an issue we’ve dealt with before. Part of the supernode/sketch structure is a pointer to where the associated data lives. Your copy constructor will need to be a little more intelligent that a direct memcpy. I can write some code for you if you want.

Ah okay, I see the pointers, I'll fix that.

CMakeLists.txt
```diff
@@ -354,10 +358,43 @@ else ()
   endif()
 endif ()
 
+add_executable(cont_test
```
etwest, line 356: So as mentioned earlier today we may want to separate these changes out into just the new verifier stuff and then a separate repo which is for the experiment code

test/cont_test.cpp
```diff
@@ -0,0 +1,102 @@
+#include <gtest/gtest.h>
```
etwest, line 1: This would be the code that would go into the separate repo along with the cmake stuff.

test/cont_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+  unsigned long num_failure = 0;
+  for (unsigned long i = 0; i < samples; i++) {
+    std::cout << "Starting updates" << std::endl;
+    for (unsigned long j = 0; j < updates_per_sample; j++) {
+      unsigned edgei = rand() % nodes;
+      unsigned edgej = rand() % (nodes - 1);
+      if (edgei > edgej) {
+        std::swap(edgei, edgej);
+      } else {
+        edgej++;
```
etwest, line 22: Could there be some weird edge case here where we do an integer overflow and end up with a self loop?
I think so. It's amazingly unlikely but you could just add an else if just in case.

test/cont_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+  std::clog << n << ',' << num_failure << std::endl;
+}
+
+//TEST(TestContinuous, TestRandom) {
+//  test_continuous(1e5, 1e7, 10);
+//}
+
+TEST(TestContinuous, DISABLED_StandardKron17) {
```
etwest, line 86: DISABLED_StandardKron17 is not very specific. Maybe NoCont_Kron17 or something?
Also, why is this here?

test/graph_test.cpp
```diff
@@ -38,15 +39,26 @@ TEST(GraphTestSuite, IFconnectedComponentsAlgRunTHENupdateLocked) {
...
+  g.set_verifier(std::make_unique<FileGraphVerifier>(curr_dir + "/res/multiples_graph_1024.txt"));
   g.connected_components();
   ASSERT_THROW(g.update({{1,2}, INSERT}), UpdateLockedException);
   ASSERT_THROW(g.update({{1,2}, DELETE}), UpdateLockedException);
 }
 
 TEST(GraphTestSuite, TestRandomGraphGeneration) {
   generate_stream();
-  GraphVerifier graphVerifier {};
+  ifstream in { "./cumul_sample.txt" };
```
etwest: What exactly is this test accomplishing? Just asserting that creating a graph verifier doesn't crash?

test/cont_test.cpp
etwest: I guess this isn't possible assuming nodes < unsigned max so never mind

CMakeLists.txt
Victor-C-Zhang: we can do that once the changes are into main

test/cont_test.cpp
Victor-C-Zhang: disabled means the test is skipped. why is it skipped?

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 cmake_minimum_required(VERSION 3.15)
 include (ExternalProject)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
Victor-C-Zhang, line 7: what functionality do we need from C++14?

CMakeLists.txt
```diff
@@ -179,14 +179,14 @@ link_directories(${CMAKE_CURRENT_BINARY_DIR}/BufferTree/prefix/lib)
...
 
-
 add_executable(tests
   test/test_runner.cpp
   test/graph_test.cpp
   test/sketch_test.cpp
   test/supernode_test.cpp
   test/util_test.cpp
-  test/util/graph_verifier.cpp
+  test/util/file_graph_verifier.cpp
```
Victor-C-Zhang, line 188: why are all these executables using both file and matrix graph verifiers? shouldn't we just pick one?

src/graph.cpp
```diff
@@ -269,11 +295,6 @@ vector<set<node_t>> Graph::parallel_connected_components() {
   return retval;
 }
 
-void Graph::post_cc_resume() {
-  GraphWorker::unpause_workers();
-  update_locked = false;
-}
-
```
Victor-C-Zhang, line 289: @etwest what's the purpose of this function? I don't see it being used in the active codebase, which is probably why Kenny's getting rid of it. At the same time, this seems like something that may have been important.

test/cont_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+  int type, a, b;
+  while (m--) {
+    in >> type >> a >> b;
+    g.update({{a, b}, INSERT});
+  }
+  g.connected_components();
+}
+
+TEST(TestContinuous, TestKron17) {
+  std::ifstream input { "./kron17" };
```
Victor-C-Zhang, line 101: maybe add a note reminding that we're assuming we have the kron graph in the runtime directory

test/graph_test.cpp
Victor-C-Zhang: A similar comment: this should probably go into a `*_graph_verifier_test` file.

test/util/mat_graph_verifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+  if (parent[i] == i) return i;
+  return parent[i] = dsu_find(parent[i], parent);
+}
+
+static void dsu_union(node_t i, node_t j, node_t* parent, node_t* size) {
+  i = dsu_find(i, parent); j = dsu_find(j, parent);
+  if (size[i] < size[j]) std::swap(i,j);
+  parent[j] = i;
+  size[i] += size[j];
+}
```
Victor-C-Zhang: nit: if this is identical to the `file_graph_verifier` functions, you could probably refactor both of them into the superclass instead.

test/util/file_graph_verifier.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <set>
+#include "../../include/supernode.h"
+#include "graph_verifier.h"
+
+/**
+ * A plugin for the Graph class that runs Boruvka alongside the graph algorithm
+ * and verifies the edges and connected components that the graph algorithm
+ * generates.
+ */
```
Victor-C-Zhang: Describe what's different between this and matrix verifier

test/util/mat_graph_verifier.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <set>
+#include "../../include/supernode.h"
+#include "graph_verifier.h"
+
+/**
+ * A plugin for the Graph class that runs Boruvka alongside the graph algorithm
+ * and verifies the edges and connected components that the graph algorithm
+ * generates.
+ */
```
Victor-C-Zhang: Describe what's different between this and file verifier

src/graph.cpp
etwest: I think it’s important to keep. Kenny is getting rid of it because his code is implementing its functionality itself. However if we write other dynamic querying code then we can use it.

maybe we can use it in Kenny’s functions so it shows up somewhere at least.

CMakeLists.txt
kennyzzhang: std::make_unique, if we want to use smart pointers it's nice to have. If c++14 is too new a standard, I can rewrite it in c++11.

src/graph.cpp
kennyzzhang: It feels like the only reason we'd need to do the `update_locked = false` operation is when we're copying the supernodes to allow requerying, which is technically not allowed by theory. The only reason we implemented this function was to do the continuous testing anyways.

test/cont_test.cpp
kennyzzhang: I didn't write this test, and I have no idea why it's here. It looks like it's just doing a normal graph verification on kron17. Shall we just remove it?

>```
> 1ce95873 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:37:57 -0700  86) TEST(TestContinuous, DISABLED_StandardKron17) {
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  87)   std::ifstream in{ "./kron17" };
> 36fa97dc test/cont_test.cpp                   (kennyzzhang       2021-10-17 16:28:20 -0400  88)   node_t n, m;
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  89)   in >> n >> m;
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  90)   Graph g{n};
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  91)   int type, a, b;
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  92)   while (m--) {
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  93)     in >> type >> a >> b;
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  94)     g.update({{a, b}, INSERT});
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  95)   }
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  96)   g.connected_components();
> cc04cd18 test/cont_test.cpp                   (Victor Zhang      2021-08-15 23:08:44 -0700  97) }
>```

test/graph_test.cpp
kennyzzhang: I have no idea what this test is accomplishing, so I'm getting rid of it.

src/graph.cpp
etwest: Future work will be in dynamic querying. For which having the ability to lock/unlock updates and pause/resume graph_workers will be essential

kennyzzhang: For dynamic querying, pass `true` to connected_components. The code has only been moved to a place where a user can't screw up resuming connected components without creating a backup copy of the supernodes.

> ```
> vector<set<node_t>> Graph::connected_components(bool cont) {
>   if (!cont)
>     return connected_components();
> 
> #ifdef USE_FBT_F
>   bf->force_flush(); // flush everything in buffertree to make final updates
> #else
>   wq->force_flush();
> #endif
> 
>   // Copy supernodes
>   Supernode** supernodes = new Supernode*[num_nodes];
>   for (node_t i=0;i<num_nodes;++i) {
>     supernodes[i] = Supernode::makeSupernode(*this->supernodes[i]);
>   }
> 
>   vector<set<node_t>> ret;
>   ret = connected_components();
>   // Restore supernodes
>   for (node_t i=0;i<num_nodes;++i) {
>     free(this->supernodes[i]);
>     this->supernodes[i] = supernodes[i];
>     representatives->insert(i);
>     parent[i] = i;
>   }
>   delete[] supernodes;
> 
>   GraphWorker::unpause_workers();
>   update_locked = false;
>   return ret;
> }
> ```

src/graph.cpp
```diff
@@ -180,12 +177,44 @@ vector<set<node_t>> Graph::connected_components() {
...
+  if (!cont)
+    return connected_components();
+
+#ifdef USE_FBT_F
+  bf->force_flush(); // flush everything in buffertree to make final updates
+#else
+  wq->force_flush();
+#endif
+
+  // Copy supernodes
```
etwest, line 194: So if you don't pause the graph workers before copying the supernodes there's no guarantee that the super nodes have the right state. They could have too few updates applied to them. This is a rare edge case but it does exist here.

test/cont_test.cpp
etwest: Sure thing

src/graph.cpp
etwest: Well this is probably not the algorithm we'll use so that complicates things a little bit. Personally, I think its cleaner to have but its also not a big deal not to have it. It's only two lines of code and can be reintroduced as necessary.

CMakeLists.txt
Victor-C-Zhang: I recall in a design discussion we had early on that we want to stick with the 11 standard

etwest: Conclusion: We decided to go with c++14

etwest (approved):
LGTM

PR #62, opened 2021-10-19
Fix unit tests so that we expect to fail 1/5000 times

etwest:
Also some small changes for statistical testing to print out the status


src/graph.cpp
```diff
@@ -211,15 +211,28 @@ vector<set<node_t>> Graph::parallel_connected_components() {
...
+    #pragma omp parallel for default(none) shared(query, reps, except, err)
     for (node_t i = 0; i < reps.size(); ++i) {
-      auto edge = supernodes[reps[i]]->sample();
+      // wrap in a try/catch because exiting through exception is undefined behavior in OMP
+      boost::optional<Edge> edge;
+      try {
+        edge = supernodes[reps[i]]->sample();
+      } catch (...) {
+        except = true;
+        err = std::current_exception();
```
Victor-C-Zhang, line 224: something to think about:
is there a case where not terminating the running thread would cause weird undefined behavior?

Victor-C-Zhang:
Allowing 2 failures per test for every test seems a bit much. Can we decrease our false positive error rate without ballooning our false negative rate too?

src/graph.cpp
etwest: Yeah it might but the idea is that the code will throw an exception any way right after the parallel loop at that point.

etwest:
> Allowing 2 failures per test for every test seems a bit much. Can we decrease our false positive error rate without ballooning our false negative rate too?

This is entirely possible but our unit test should be quickly alerting us to “this code is very wrong” whereas the statistical testing is responsible for a more fine grained analysis.

Basically this means that a false positive is much worse in this context than a false negative. With say 1 allowed failure per test we end up failing 13 times in 1000 tests.

Victor-C-Zhang:
> > Allowing 2 failures per test for every test seems a bit much. Can we decrease our false positive error rate without ballooning our false negative rate too?
> 
> This is entirely possible but our unit test should be quickly alerting us to “this code is very wrong” whereas the statistical testing is responsible for a more fine grained analysis.
> 
> Basically this means that a false positive is much worse in this context than a false negative. With say 1 allowed failure per test we end up failing 13 times in 1000 tests.

Ok. That makes sense. Since you're doing this 2/10 tolerance a bunch of times, is there a way to wrap this all in a function? If not, can you add a little blurb at the top of the file describing the above rationale?

test/parallel_graph_test.cpp
```diff
@@ -3,6 +3,16 @@
 #include "../include/graph.h"
 #include "util/graph_gen.h"
 
+/**
+ * For many of these tests (especially for those upon very sparse and small graphs)
+ * we allow for a certain number of failures per test.
+ * This is because the responsibility of these tests is to quickly alert us 
+ * to “this code is very wrong” whereas the statistical testing is responsible 
+ * for a more fine grained analysis.
+ * In this context a false false positive is much worse than a false negative.
```
Victor-C-Zhang: “false false”

test/graph_test.cpp
```diff
@@ -4,6 +4,16 @@
 #include "util/graph_verifier.h"
 #include "util/graph_gen.h"
 
+/**
+ * For many of these tests (especially for those upon very sparse and small graphs)
+ * we allow for a certain number of failures per test.
+ * This is because the responsibility of these tests is to quickly alert us 
+ * to “this code is very wrong” whereas the statistical testing is responsible 
+ * for a more fine grained analysis.
+ * In this context a false false positive is much worse than a false negative.
```
Victor-C-Zhang: “false false”

Victor-C-Zhang (approved):
Approved, delta small typo fix

PR #65, opened 2021-10-28
Changes for Buffering Repo Rename+Standalone

etwest:
We've moved the standalone gutters to the buffering repo and we've renamed some elements. This pull request adapts the streaming code to these changes.


etwest:
Last thing to do is to switch from having a unique executable for gutter tree vs standalone to just have a streaming.conf option for it.

etwest:
Ready for review

etwest:
Reminder to myself to switch to main branch of GutterTree repo once that pull request is in

src/graph.cpp
```diff
@@ -124,74 +113,7 @@ void Graph::batch_update(uint64_t src, const std::vector<uint64_t>& edges, Super
...
-    modified = false;
-    vector<node_t> removed;
-    for (node_t i: (*representatives)) {
-      if (parent[i] != i) continue;
-      boost::optional<Edge> edge = supernodes[i]->sample();
-#ifdef VERIFY_SAMPLES_F
-      if (edge.is_initialized())
-        verifier.verify_edge(edge.value());
-      else
-        verifier.verify_cc(i);
```
etwest: So, currently the parallel connected components algorithm doesn't have this type of verification? Is this something we want? And is it trivial to add?

Victor-C-Zhang: I didn’t implement this in parallel because I assumed it was always going to be secondary to the single-threaded implementation. I think it won’t be too difficult to modify the impl of edge verify to wait until all queries to update CCs. I think verification is necessary to check correctness.

CMakeLists.txt
```diff
@@ -146,16 +146,16 @@ if(NOT Boost_FOUND)
...
+# Install GutterTree Project
 ExternalProject_Add(
-  BufferTree
+  GutterTree
 
-  PREFIX          BufferTree
+  PREFIX          GutterTree
   GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/FastBufferTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "async_3_electric-er"
```
Victor-C-Zhang: make sure to change back!

.github/workflows/cmake.yml
```diff
@@ -87,27 +87,9 @@ jobs:
         # Execute the build.  You can specify a specific target with "--target <NAME>"
         run: cmake --build . --config $BUILD_TYPE
```
Victor-C-Zhang, line 89: nit: maybe we want to make these matrix builds?

example_streaming.conf
```diff
@@ -1,5 +1,8 @@
+# Should we use the GutterTree as the buffering system or the standalone gutters ("tree" for GutterTree, "gutters" for standalone gutters)
```
Victor-C-Zhang: “standalone” instead of “gutters” imo

.github/workflows/cmake.yml
etwest: For multiple OS?

src/graph.cpp
etwest: Done

CMakeLists.txt
etwest: Will do

example_streaming.conf
etwest: Yeah that seems reasonable

etwest: Done

include/test/mat_graph_verifier.h
```diff
@@ -28,4 +28,27 @@ class MatGraphVerifier : public GraphVerifier {
    * @return an array of connected components.
    */
   static std::vector<std::set<node_t>> kruskal(node_t n, const std::vector<bool>& input);
+
+  // Returns a unique identifier for each edge using diagonalization
+  static inline uint64_t get_uid(node_t i, node_t j) {
```
Victor-C-Zhang, line 33: is this any different from what we're doing in `util.cpp`?

etwest: Yea, the one in util.cpp is now optimized for speed while this one is optimized for space

etwest: (The one in util just concatenates the node_ids to get an edge_id)

Victor-C-Zhang: The one in `util.cpp` causes our data structure to be slightly larger then right? It's probably just `n(log^3(n^2) - log^3(n^2-n))` so not the end of the world.

Victor-C-Zhang (approved):
Approving delta changing branch of gutter tree to `main`

include/test/mat_graph_verifier.h
etwest: It does indeed but not when node ids are within 32bits. This might be something to revisit when we go up to 64 bit ids.

PR #66, opened 2021-10-29
Sketch Refactor log^3 bits

etwest:
We realized that our implementation has higher asymptotic data cost then it needs to. The changes in this pull request fix this issue.

### DIFF
--- a/include/l0_sampling/sketch.h
+++ b/include/l0_sampling/sketch.h
@@ -6,13 +6,21 @@
 #include <vector>
 #include <memory>
 #include <mutex>
+#include <utility>
 #include "../bucket.h"
 #include "../types.h"
 #include "../util.h"
 #include <gtest/gtest_prod.h>
 
+// #define bucket_gen(d) double_to_ull((log2(d)+1))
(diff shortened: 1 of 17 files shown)


include/l0_sampling/sketch.h
```diff
@@ -50,30 +57,30 @@ class Sketch {
    * Construct a sketch of a vector of size n
    * @param n Length of the vector to sketch. (static variable)
    * @param seed Seed to use for hashing operations
-   * @param num_bucket_factor Factor to scale the number of buckets in this sketch (static variable)
+   * @param failure_factor The rate at which an individual sketch is allowed to fail (determines column width)
```
Victor-C-Zhang: besides seed, these aren't params for the constructor. put them where they are used (maybe config?)

Victor-C-Zhang: Either that or make it clear that these are the list of possible params for all constructors/factory methods

etwest: Should be done now

Victor-C-Zhang (approved):

PR #68, opened 2021-11-09
refactor: import graph and sketch types from common repo

### LINKED ISSUE #57: Remove dependence on FastBufferTree Node type
The two types `Node` in FastBufferTree and `node_t` in StreamingGraphAlgo represent the same type, but are named differently. This is probably not a good idea. The ideal solution is to have a "common" components package that both repos consume these types (among other things). We should have a design discussion about this at some point in the future.

Victor-C-Zhang:
Closes #57


CMakeLists.txt
```diff
@@ -146,13 +146,33 @@ if(NOT Boost_FOUND)
 endif()
 include_directories("${Boost_INCLUDE_DIR}")
 
+# Include common components
+ExternalProject_Add(
+    GraphZeppelinCommon
+
+    PREFIX          GraphZeppelinCommon
+    GIT_REPOSITORY  "https://github.com/Victor-C-Zhang/GraphZeppelinCommon.git"
```
Victor-C-Zhang: move repo to org

etwest (approved):
looks good

PR #76, opened 2022-01-17
Remove dependence upon boost

etwest:
Boost is annoying and we don't need it.

Feel free to argue with me on this one. I'm feeling a little bit spiteful.


Victor-C-Zhang:
Why did we stop returning `optional`s for `Supernode::sample()`?

etwest:
> Why did we stop returning `optional`s for `Supernode::sample()`?

Once we made the change for sketches to allow a constant failure probability for each sketch we decided to move away from throwing exceptions to communicate zero and sketch failure. A natural consequence was that we needed to return a code to indicate this along with the edge so it was no longer necessary to use a optional.

We stopped using exceptions because they're ugly and slow :)

Victor-C-Zhang (approved):

PR #79, opened 2022-02-02
Perf improve

etwest:
Seems to make the code marginally faster. I'll post results here in a bit

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,7 @@ target_link_libraries(GraphStreamingCC PUBLIC xxhash GutterTree)
 target_include_directories(GraphStreamingCC PUBLIC include/ include/l0_sampling/)
 target_compile_options(GraphStreamingCC PUBLIC -fopenmp)
 target_link_options(GraphStreamingCC PUBLIC -fopenmp)
+target_compile_definitions(GraphStreamingCC PUBLIC XXH_INLINE_ALL)
 
 add_library(GraphStreamingVerifyCC
   src/graph.cpp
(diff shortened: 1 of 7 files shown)


etwest:
Results of unlimited speed experiment run with buffer size of -4

kron13  -- 4.60 million per second  -- 0.11 second CC
kron15  -- 4.11 million per second -- 0.53 second CC
kron16  -- 3.84 million per second -- 1.22 second CC
kron17  -- 3.58 million per second -- 2.76 second CC
kron18  -- 3.46 million per second -- 5.99 second CC

CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ FetchContent_Declare(
   GutterTree
 
   GIT_REPOSITORY  "https://github.com/GraphStreamingProject/GutterTree.git"
-  GIT_TAG         "main"
+  GIT_TAG         "word_size_fix"
```
Victor-C-Zhang: yktw

include/bucket.h
```diff
@@ -61,12 +61,8 @@ namespace Bucket_Boruvka {
...
 } // namespace Bucket_Boruvka
 
-inline col_hash_t Bucket_Boruvka::col_index_hash(const unsigned bucket_col, const vec_t& update_idx, const long sketch_seed) {
-  struct {
-    unsigned bucket_col;
-    vec_t update_idx;
-  } __attribute__((packed)) buf = {bucket_col, update_idx};
-  return col_hash(&buf, sizeof(buf), sketch_seed);
+inline col_hash_t Bucket_Boruvka::col_index_hash(const vec_t& update_idx, const long seed_and_col) {
+  return col_hash(&update_idx, sizeof(update_idx), seed_and_col);
```
Victor-C-Zhang, line 65: is this the main change?

etwest: This and the node_id_t -> size_t

Victor-C-Zhang: do u mind summarizing what this change does?

etwest: So basically, instead of seeding by appending the sketch column to the beginning of the data we modify the sketch seed itself.

This reduces the size of the data and saves us the time to construct the packed data.

Victor-C-Zhang (approved):

PR #81, opened 2022-02-04
Fixed matrix verifier to actually work

etwest:
Title 

Also I'm going to add some time hooks.

Also also, some small changes to the way we test sketches. We expect a sketch failure rate of 1/100 and the tests did not take this into account

### DIFF
--- a/include/test/mat_graph_verifier.h
+++ b/include/test/mat_graph_verifier.h
@@ -12,43 +12,24 @@
 class MatGraphVerifier : public GraphVerifier {
   std::vector<std::set<node_id_t>> kruskal_ref;
   std::vector<std::set<node_id_t>> boruvka_cc;
-  std::vector<bool>& det_graph;
+  std::vector<std::vector<bool>> adj_graph;
   DisjointSetUnion<node_id_t> sets;
 
-public:
-  MatGraphVerifier(node_id_t n, std::vector<bool>& input);
-
-  void verify_edge(Edge edge);
-  void verify_cc(node_id_t node);
-  void verify_soln(std::vector<std::set<node_id_t>> &retval);
+  node_id_t n;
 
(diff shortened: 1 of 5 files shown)


include/test/mat_graph_verifier.h
```diff
@@ -12,43 +12,24 @@
 class MatGraphVerifier : public GraphVerifier {
   std::vector<std::set<node_id_t>> kruskal_ref;
   std::vector<std::set<node_id_t>> boruvka_cc;
-  std::vector<bool>& det_graph;
+  std::vector<std::vector<bool>> adj_graph;
```
Victor-C-Zhang, line 15: is this the fix?

Victor-C-Zhang (commented):
If having this merged isn't critical for submission, we should merge this and the time_hook branch

include/test/mat_graph_verifier.h
etwest: It is most of it. The idea is to avoid the edge_id map and just use an adjacency matrix.

etwest:
> If having this merged isn't critical for submission, we should merge this and the time_hook branch

What's in there that we want?

Victor-C-Zhang:
> > If having this merged isn't critical for submission, we should merge this and the time_hook branch
> 
> What's in there that we want?

A lot of the same time hooks you have that the current experiments rely on

etwest:
> > > If having this merged isn't critical for submission, we should merge this and the time_hook branch
> > 
> > 
> > What's in there that we want?
> 
> A lot of the same time hooks you have that the current experiments rely on

I've got those in here as well. Can we merge this and then open another pull request for anything we want from the time_hook branch?

Victor-C-Zhang (approved):

PR #82, opened 2022-02-08
feat: time hooks for query experiment

Victor-C-Zhang:
- Merge time hooks from continuous query branch
- Ensure supernodes are restored if a continuous query errors out
- Bugfixes and improvements:
  - Reversed order of parameters in `merge_supernodes`
  - Use `const&` references for vectors
  - Improve robustness of `sizeof` calculations in binary IO operations
  - etc.


include/graph.h
```diff
@@ -101,8 +101,9 @@ class Graph {
    */
   void write_binary(const std::string &filename);
 
-  std::chrono::steady_clock::time_point flush_call;
-  std::chrono::steady_clock::time_point flush_return;
+  // time hooks for experiments
+  std::chrono::steady_clock::time_point flush_start;
```
etwest, line 157: Why did we rename these

src/graph.cpp
```diff
@@ -217,7 +221,7 @@ std::vector<std::set<node_id_t>> Graph::boruvka_emulation() {
   for (const auto& it : temp) retval.push_back(it.second);
 
   cc_alg_end = std::chrono::steady_clock::now();
-  printf("CC done\n");
+  printf("CC done in %d rounds\n", num_rounds);
```
etwest: Probably not necessary

test/simulation.cpp
```diff
@@ -0,0 +1,43 @@
+#include <random>
```
etwest: What exactly is this simulation? A stress test of the std priority queue?

test/temp.cpp
```diff
@@ -0,0 +1,28 @@
+#include <iostream>
```
etwest: What is this file?

test/res/kron17_trouble.txt
```diff
@@ -0,0 +1,9 @@
+131072  8
```
etwest: What is this file for?

test/simulation.cpp
Victor-C-Zhang: oops, an unnecessary

test/temp.cpp
Victor-C-Zhang: same

test/res/kron17_trouble.txt
Victor-C-Zhang: same

src/graph.cpp
Victor-C-Zhang: I used this for some experiments, but I'm not too attached to it if you don't like it

include/graph.h
Victor-C-Zhang: for consistency

etwest:
This looks good. I'll get Ethan's changes in for splitting cc up into multiple functions and then I'll push a commit here to ensure this is done correctly once those changes are made.

include/graph.h
etwest: Okay

etwest: Okay

src/graph.cpp
etwest: Yeah lets get rid of it. We can always add it back if we want to. Also, the cc done printf isn't necessary either.

include/graph.h
```diff
@@ -1,165 +1,162 @@
...
+class Graph {
+  node_id_t num_nodes;
+  long seed;
+  bool update_locked = false;
+  bool modified = false;
+  // a set containing one "representative" from each supernode
+  std::set<node_id_t>* representatives;
+  Supernode** supernodes;
+  // DSU representation of supernode relationship
+  node_id_t * parent;
```
etwest: Style nit: Remove space after *

include/graph.h
etwest: Can you put in a pull request for the experiment repo that changes the query_expr in main to use these time hooks?

src/graph.cpp
```diff
@@ -186,25 +186,25 @@ inline std::vector<std::vector<node_id_t>> Graph::supernodes_to_merge(std::pair<
...
 inline void Graph::merge_supernodes(Supernode** copy_supernodes, std::vector<node_id_t> &new_reps,
-               std::vector<std::vector<node_id_t>> &to_merge, bool make_copy, bool first_round) {
+               std::vector<std::vector<node_id_t>> &to_merge, bool first_round, bool make_copy) {
   bool except = false;
   std::exception_ptr err;
   // loop over the to_merge vector and perform supernode merging
   #pragma omp parallel for default(shared)
-  for (node_id_t i = 0; i < new_reps.size(); i++) {
-    node_id_t a = new_reps[i]; // fix for ubuntu18.04 ('for each' style in omp no good)
+  for (node_id_t i = 0; i < new_reps.size(); i++) { // NOLINT(modernize-loop-convert)
```
etwest, line 206: What exactly does this comment mean?

src/graph.cpp
```diff
@@ -186,25 +186,25 @@ inline std::vector<std::vector<node_id_t>> Graph::supernodes_to_merge(std::pair<
...
-    if (to_merge[a].size() != 0) new_reps.push_back(a);
+    if (!to_merge[a].empty()) new_reps.push_back(a);
 
   reps = new_reps;
   return to_merge;
 }
 
 inline void Graph::merge_supernodes(Supernode** copy_supernodes, std::vector<node_id_t> &new_reps,
-               std::vector<std::vector<node_id_t>> &to_merge, bool make_copy, bool first_round) {
+               std::vector<std::vector<node_id_t>> &to_merge, bool first_round, bool make_copy) {
```
etwest: Nice catch. That might have explained part of the difference in performance we were observing.

src/graph.cpp
```diff
@@ -242,21 +243,42 @@ std::vector<std::set<node_id_t>> Graph::boruvka_emulation(bool make_copy) {
...
-    modified = false;
-    sample_supernodes(query, reps);
-    std::vector<std::vector<node_id_t>> to_merge = supernodes_to_merge(query, reps);
-
-    // make a copy if necessary
-    if(make_copy && first_round) {
-      backed_up = reps;
-      if (!copy_in_mem) backup_to_disk(backed_up);
+  // function to restore supernodes after CC if make_copy is specified
+  auto cleanup_copy = [&make_copy, this, &backed_up, &copy_supernodes]() {
```
etwest, line 248: This code looks right. I'm curious if you've tried inducing an exception to ensure the super-nodes are restored correctly. Maybe this should be a unit test?

src/graph.cpp
etwest: Actually, semantically it probably doesn't make a difference. I think we can (and should) reduce the number of arguments to this function to just one: namely 'make_copy' and pass in whether we require the supernodes to be backed up. (first_round && make_copy) or just first_round which starts true only if make_copy is true.

src/graph.cpp
```diff
@@ -313,17 +323,26 @@ void Graph::restore_from_disk(std::vector<node_id_t> ids_to_restore) {
...
-  flush_call = std::chrono::steady_clock::now();
+  flush_start = std::chrono::steady_clock::now();
   bf->force_flush(); // flush everything in buffering system to make final updates
   GraphWorker::pause_workers(); // wait for the workers to finish applying the updates
-  flush_return = std::chrono::steady_clock::now();
+  flush_end = std::chrono::steady_clock::now();
   // after this point all updates have been processed from the buffer tree
 
-  if (!cont)
+  if (!cont) {
```
etwest: style nit: {} unnecessary

src/graph.cpp
```diff
@@ -348,10 +370,10 @@ void Graph::write_binary(const std::string& filename) {
   // after this point all updates have been processed from the buffering system
 
   auto binary_out = std::fstream(filename, std::ios::out | std::ios::binary);
-  int fail_factor = Sketch::get_failure_factor();
-  binary_out.write((char*)&seed, sizeof(long));
-  binary_out.write((char*)&num_nodes, sizeof(uint64_t));
-  binary_out.write((char*)&fail_factor, sizeof(int));
+  auto fail_factor = Sketch::get_failure_factor();
```
etwest, line 373: This makes me nervous. I suppose if it works then it works though.

include/graph.h
Victor-C-Zhang: Yeah, it’s in draft rn

src/graph.cpp
Victor-C-Zhang: It’s a directive for CLion so it doesn’t mark this as a loop that should be modernized

src/graph.cpp
Victor-C-Zhang: Yes, you’re right. To do so we’ll have to add a dependency that allows us to mock. I can work on that if you think it’s important.

src/graph.cpp
Victor-C-Zhang: what part of it?

src/graph.cpp
etwest: Could you add the comment below back in. I want to explain why we aren't using the modern style.

src/graph.cpp
etwest: Using auto for a variable we're going about to write to disk. I guess I'm nervous the compiler might chose some weird larger or smaller type to put fail_factor in. Though I don't know if that's actually possible.

src/graph.cpp
etwest: I'll add it to the Trello and either you or someone else can be assigned to it. Doesn't have to be a part of this PR.

src/graph.cpp
Victor-C-Zhang: auto takes the type of the RHS at compile time. it’s like if we looked up the type of get_failure_factor (uint32) and spliced it in

src/graph.cpp
Victor-C-Zhang: sure, ill make that change

src/graph.cpp
etwest: Gotcha, sounds good

etwest (approved):
LGTM

PR #83, opened 2022-02-08
Improvements to Connected Components Performance

etwest:
## Big Changes
- We now can either backup supernodes to disk or in memory based upon a configuration flag set by the user
- Made Boruvka emulation much more parallel by serially computing the DSU changes and tracking the merges that will need to happen. Then in parallel performing the merges. The boruvka emulation now consists of the following three stages performed repeatedly.
  1. Query Active Supernodes (parallel)
  2. Perform DSU updates and identify merging to be done (serial)
  3. Merge the Supernodes (parallel)
- Additionally, changed the backing up of supernodes to only occur when the supernode in question has been selected by the DSU stage for merging during the first round of Boruvka

#### Configuration
- Renamed 'path_prefix' in configuration options to 'disk_dir' to make it clear what it does.
- 'disk_dir' is now where both the gutter tree and Supernodes backed up to disk live

#### Connected Components Algorithm
- Made Graph::update inline (maybe minor performance benefit)
- Added ability to reset Supernode idx and Sketch already_queried variables (useful for continuous CC)
- Renamed function that runs the CC algorithm to 'boruvka_emulation'
- Made all CCs route through Graph::connected_components(bool) and made that the only function that calls BufferingSystem::force_flush()

#### Sketching
- Fixed bug in sketching (we need to use 64 bit column hashes as (2^32)^2 = 2^64)
- Changed contains query to use bitwise & rather than modulus as modulus would break on 2^32 nodes (as we would need to check if all 64 bits are 1 and therefore would attempt to mod by 2^65.

#### Testing
- Fixes graph tests to not allow for failures as these failures should not happen whp
- Add a test for queries during the stream that backs up in memory and on disk
- Add a test that uses parallel processing

### DIFF
--- a/include/bucket.h
+++ b/include/bucket.h
@@ -67,11 +67,10 @@ inline col_hash_t Bucket_Boruvka::col_index_hash(const vec_t& update_idx, const 
 
 inline vec_hash_t Bucket_Boruvka::index_hash(const vec_t& index, long sketch_seed) {
   return vec_hash(&index, sizeof(index), sketch_seed);
-
 }
 
-inline bool Bucket_Boruvka::contains(const col_hash_t& col_index_hash, const vec_t& guess_nonzero) {
-  return col_index_hash % guess_nonzero == 0;
+inline bool Bucket_Boruvka::contains(const col_hash_t& col_index_hash, const col_hash_t& guess_nonzero) {
+  return (col_index_hash & guess_nonzero) == 0;
 }
 
 inline bool Bucket_Boruvka::is_good(const vec_t& a, const vec_hash_t& c, const long& sketch_seed) {
--- a/include/bucket.h
+++ b/include/bucket.h
@@ -26,7 +26,7 @@ namespace Bucket_Boruvka {
   /**
    * Checks whether the hash associated with the Bucket hashes the index to 0.
    * @param col_index_hash The return value to Bucket::col_index_hash
-   * @param guess_nonzero A power of 2, used as a modulus
+   * @param guess_nonzero A power of 2, used to check if a specific bit value is non-zero
    * @return true if the index is NOT hashed to zero mod guess_nonzero.
    */
   inline static bool contains(const col_hash_t& col_index_hash, const vec_t& guess_nonzero);
@@ -70,7 +70,7 @@ inline vec_hash_t Bucket_Boruvka::index_hash(const vec_t& index, long sketch_see
 }
 
 inline bool Bucket_Boruvka::contains(const col_hash_t& col_index_hash, const col_hash_t& guess_nonzero) {
-  return (col_index_hash & guess_nonzero) == 0;
+  return (col_index_hash & guess_nonzero) == 0; // use guess_nonzero (power of 2) to check ith bit
 }
 
 inline bool Bucket_Boruvka::is_good(const vec_t& a, const vec_hash_t& c, const long& sketch_seed) {
--- a/include/l0_sampling/sketch.h
+++ b/include/l0_sampling/sketch.h
@@ -41,7 +41,7 @@ private:
   vec_hash_t* bucket_c;
 
   // Flag to keep track if this sketch has already been queried.
-  bool already_quered = false;
+  bool already_queried = false;
 
   FRIEND_TEST(SketchTestSuite, TestExceptions);
   FRIEND_TEST(EXPR_Parallelism, N10kU100k);
@@ -96,6 +96,10 @@ public:
   
   inline static vec_t get_failure_factor() 
   { return failure_factor; }
+
+  inline void reset_queried() 
+  { already_queried = false; }
+
   /**
    * Update a sketch based on information about one of its indices.
    * @param update the point update.
--- a/include/test/write_configuration.h
+++ b/include/test/write_configuration.h
@@ -1,16 +1,17 @@
 #include <fstream>
 
-static void write_configuration(bool use_tree, int groups=1, int g_size=1) {
+static void write_configuration(bool use_tree, bool backup_in_mem=false, int groups=1, int g_size=1) {
 	// read previous configuration to get GutterTree prefix
 	// as this is system dependent and shouldn't be set by our code
 	std::ifstream in("streaming.conf");
-	std::string path_prefix;
+	std::string disk_dir = ".";
 	if (in.is_open()) {
 		std::string line;
 		while(getline(in, line)) {
 			if (line[0] == '#' || line[0] == '\n') continue;
-			if (line.substr(0, line.find('=')) == "path_prefix") {
-				path_prefix = line.substr(line.find('=') + 1);
+			if (line.substr(0, line.find('=')) == "disk_dir") {
+				disk_dir = line.substr(line.find('=') + 1);
+				if (disk_dir == "") disk_dir = ".";
 				break;
 			}
 		}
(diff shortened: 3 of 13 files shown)


test/graph_test.cpp
```diff
@@ -181,3 +150,96 @@ TEST_P(GraphTest, TestCorrectnessOfReheating) {
     }
   }
 }
+
+TEST_P(GraphTest, MultipleInserters) {
```
etwest, line 156: Does anyone know what this test is? I'm not sure how it got in here

Victor-C-Zhang: you added this with commit "on demand copying of supernodes to disk"

include/bucket.h
```diff
@@ -67,11 +67,10 @@ inline col_hash_t Bucket_Boruvka::col_index_hash(const vec_t& update_idx, const
 
 inline vec_hash_t Bucket_Boruvka::index_hash(const vec_t& index, long sketch_seed) {
   return vec_hash(&index, sizeof(index), sketch_seed);
-
 }
 
-inline bool Bucket_Boruvka::contains(const col_hash_t& col_index_hash, const vec_t& guess_nonzero) {
-  return col_index_hash % guess_nonzero == 0;
+inline bool Bucket_Boruvka::contains(const col_hash_t& col_index_hash, const col_hash_t& guess_nonzero) {
```
Victor-C-Zhang, line 72: can you add a comment either here or where `guess_nonzero` is set that it needs to be a power of 2 to allow us to use this hack

test/graph_test.cpp
etwest: Oof. Okay I'll get rid of it

include/bucket.h
etwest: Sure thing

test/graph_test.cpp
etwest: Oh! Actually, I do remember what this test is. It's testing the multithreading by setting the number of Graph Workers and the size of each group to higher values. I'll add a comment explaining its purpose

include/l0_sampling/sketch.h
```diff
@@ -41,7 +41,7 @@ class Sketch {
   vec_hash_t* bucket_c;
 
   // Flag to keep track if this sketch has already been queried.
-  bool already_quered = false;
+  bool already_queried = false;
```
Victor-C-Zhang, line 44: lol

include/test/write_configuration.h
```diff
@@ -1,16 +1,17 @@
 #include <fstream>
 
-static void write_configuration(bool use_tree, int groups=1, int g_size=1) {
+static void write_configuration(bool use_tree, bool backup_in_mem=false, int groups=1, int g_size=1) {
 	// read previous configuration to get GutterTree prefix
 	// as this is system dependent and shouldn't be set by our code
 	std::ifstream in("streaming.conf");
-	std::string path_prefix;
+	std::string disk_dir = ".";
```
Victor-C-Zhang, line 7: don't know if this is a good idea. end user may be unpleasantly surprised to discover a giant file on their shiny new machine if they aren't fully aware of our system

src/graph.cpp
```diff
@@ -91,15 +95,6 @@ Graph::~Graph() {
   open_graph = false;
 }
 
-void Graph::update(GraphUpdate upd) {
```
Victor-C-Zhang, line 94: does inlining this make us faster?

src/graph.cpp
```diff
@@ -216,33 +256,50 @@ std::vector<std::set<node_id_t>> Graph::boruvka_emulation() {
   retval.reserve(temp.size());
   for (const auto& it : temp) retval.push_back(it.second);
 
+  if (make_copy) {
+    if(copy_in_mem) {
+      // restore original supernodes and free memory
+      for (node_id_t i = 0; i < num_nodes; i++) {
+        if (supernodes[i] != nullptr) free(supernodes[i]);
```
Victor-C-Zhang, line 263: why do we need to check this?

src/graph.cpp
```diff
@@ -216,33 +256,50 @@ std::vector<std::set<node_id_t>> Graph::boruvka_emulation() {
...
+        supernodes[i] = copy_supernodes[i];
```
Victor-C-Zhang, line 259: why can't we just put `supernodes = copy_supernodes`?

src/l0_sampling/sketch.cpp
```diff
@@ -55,13 +55,13 @@ Sketch::Sketch(const Sketch& s) : seed(s.seed) {
...
 void Sketch::update(const vec_t& update_idx) {
-  XXH64_hash_t update_hash = Bucket_Boruvka::index_hash(update_idx, seed);
+  vec_hash_t update_hash = Bucket_Boruvka::index_hash(update_idx, seed);
   Bucket_Boruvka::update(bucket_a[num_elems - 1], bucket_c[num_elems - 1], update_idx, update_hash);
   for (unsigned i = 0; i < num_buckets; ++i) {
     col_hash_t col_index_hash = Bucket_Boruvka::col_index_hash(update_idx, seed + i);
     for (unsigned j = 0; j < num_guesses; ++j) {
       unsigned bucket_id = i * num_guesses + j;
-      if (Bucket_Boruvka::contains(col_index_hash, 2 << j)){
+      if (Bucket_Boruvka::contains(col_index_hash, ((col_hash_t)1) << j)){
```
Victor-C-Zhang, line 64: why do we change from 2^{j-1} to 2^j?

include/l0_sampling/sketch.h
etwest: :)

src/graph.cpp
etwest: Just ran a test. Seems to mostly be a difference for small graphs.

| Dataset | Inlining update | No inlining |
|-|-|-|
| kron13 | 5.2698 | 5.1689 |
| kron15 | 4.8033 | 4.7994 |

include/test/write_configuration.h
etwest: In this case this is for the unit tests so this file size will be relatively small (probably around a couple MiB)

Are you talking more generally about the disk_dir option?

src/graph.cpp
etwest: Because some of the copy_supernodes are nullptrs and because we want to loop through to free the memory anyway. This was super rushed though so I'm welcome to better ideas

src/l0_sampling/sketch.cpp
etwest: Because the contains query changed from `hash % 2^x == 0` to `hash & 2^x == 0`. In the first version we need j to be one larger than the index we're querying while in the second we don't

include/test/write_configuration.h
Victor-C-Zhang: I'm saying in general. this is the same code that we use when running experiments right?

src/graph.cpp
Victor-C-Zhang: ah gotcha. i think it would be a better idea to follow the same pattern for in-memory as we do on disk. that is, copy the sketches we will modify to storage and restore those sketches rather than working on the copy and pointing all the supernodes to the new-but-also-not-new sketches.

include/test/write_configuration.h
etwest: I believe it's basically the same. Actually something we probably should do at some point is to make them literally the same.

etwest: What do you think the solution is? We do need to use disk.

Victor-C-Zhang: make the user provide a `disk_dir` rather than assuming a default

etwest: I think its more annoying to force the user to specify a configuration file when they may just want to run our system on a random graph and see what happens. However, I think your point is valid. So I'll change the setup to default to using the standalone gutters and backing up in memory. That way in order for our system to use disk the user needs to specify a configuration.

Sound good?

Victor-C-Zhang: good to me

src/graph.cpp
etwest: Took a stab at doing this in the second most recent commit

Victor-C-Zhang (approved):
lgtm

PR #87, opened 2022-02-27
Benchmarking script

etwest:
Added optional benchmarks for a variety of GraphStreamingCC's features.

#### Current Benchmarks
* BinaryGraphStream throughput
* Sketch Update Rate
* Sketch Query Rate
* Hashing speed

I'm open to ideas for more benchmarks to add to this pull request.


bench/graphcc_bench.cpp
```diff
@@ -0,0 +1,180 @@
...
+// batches by seed
+// The argument to this benchmark is the number of hashes to batch
+static void BM_Hash_XXH64(benchmark::State &state) { 
+  uint64_t num_seeds = 8;
+  uint64_t num_hashes = state.range(0);
+  uint64_t output;
+  for (auto _ : state) {
+    for (uint64_t h = 0; h < num_seeds; h++) {
+      for (uint64_t i = 0; i < num_hashes; i++) {
+        benchmark::DoNotOptimize(output = XXH64(&i, sizeof(uint64_t), seed + h));
```
Victor-C-Zhang: why no optimization?

Victor-C-Zhang (commented):
I think the actual benchmarking code should go in our experiments repo. We aren’t going to be running it for every PR right?

bench/graphcc_bench.cpp
etwest: So benchmark::DoNotOptimize() prevents the results of the expression from being optimized out. Optimizations still happen within the expression ie XXH64 still gets optimized but the result is forced to be written to output.

etwest: That prevents the compiler from going "Oh you never actually use 'output'. Let me just optimize out all of this work and make the benchmark a no-op"

etwest:
> I think the actual benchmarking code should go in our experiments repo. We aren’t going to be running it for every PR right?

The benchmarks only get built if the BUILD_BENCH flag is passed to the cmake command so they wouldn't be automatically run for our GitHub Actions or included when building as an external project.
My thinking is that since the benchmarks test the performance of the library they have a place here.

Victor-C-Zhang:
> > I think the actual benchmarking code should go in our experiments repo. We aren’t going to be running it for every PR right?
> 
> The benchmarks only get built if the BUILD_BENCH flag is passed to the cmake command so they wouldn't be automatically run for our GitHub Actions or included when building as an external project. My thinking is that since the benchmarks test the performance of the library they have a place here.

Do we also have the once-a-week correctness tests in this repo too?

etwest:
> > > I think the actual benchmarking code should go in our experiments repo. We aren’t going to be running it for every PR right?
> > 
> > 
> > The benchmarks only get built if the BUILD_BENCH flag is passed to the cmake command so they wouldn't be automatically run for our GitHub Actions or included when building as an external project. My thinking is that since the benchmarks test the performance of the library they have a place here.
> 
> Do we also have the once-a-week correctness tests in this repo too?

Yeah, although those don’t really work anymore. Probably something to put on the list of things to work on.

Victor-C-Zhang:
> > > > I think the actual benchmarking code should go in our experiments repo. We aren’t going to be running it for every PR right?
> > > 
> > > 
> > > The benchmarks only get built if the BUILD_BENCH flag is passed to the cmake command so they wouldn't be automatically run for our GitHub Actions or included when building as an external project. My thinking is that since the benchmarks test the performance of the library they have a place here.
> > 
> > 
> > Do we also have the once-a-week correctness tests in this repo too?
> 
> Yeah, although those don’t really work anymore. Probably something to put on the list of things to work on.

Maybe put these benchmarks in with those benchmarking files as well and we'll call it a day.

etwest:
The correctness tests are a test though not a benchmark. Do you not like the bench directory? Maybe I can make it a sub directory of tests?

Victor-C-Zhang:
> The correctness tests are a test though not a benchmark. Do you not like the bench directory? Maybe I can make it a sub directory of tests?

But the purposes are the same right? We run this every so often to make sure nothing’s going wrong

etwest:
> > The correctness tests are a test though not a benchmark. Do you not like the bench directory? Maybe I can make it a sub directory of tests?
> 
> But the purposes are the same right? We run this every so often to make sure nothing’s going wrong

I see, I was imagining these being used for when we're trying to improve the performance of the algorithm rather than having to run the speed_experiments and for identifying the bottlenecks in our system. I suppose they could also be used for checking that we haven't made changes that have a detrimental affect on performance.

Victor-C-Zhang (dismissed):

Victor-C-Zhang (approved):

PR #89, opened 2022-03-02
Automatically update repos we depend on

etwest:
Previously running `cmake ..` would not pull in updates from our other repos. This small change fixes this issue.


Victor-C-Zhang:
was it the quotation marks or the update disconnection?

Victor-C-Zhang (approved):

etwest:
> was it the quotation marks or the update disconnection?

Quotes. Those other options didn’t actually do anything.

PR #91, opened 2022-03-09
Fix configuration to actually allow for guttertree

etwest:
In process of fixing the bug Victor found last night I discovered that we weren't allowing for running with the gutter tree. Oops.


etwest:
Once the changes in the GutterTree repo are merged these tests will pass.

Victor-C-Zhang (approved):
Wait huh. So all of our experiments were a lie?

etwest:
> Wait huh. So all of our experiments were a lie?

No definitely not. This problem was introduced post experiments for the paper.

PR #92, opened 2022-03-10
General write_binary functions

etwest:
Use std::ostream and std::istream to allow for general purpose writing of supernodes and sketches to binary either in a file or in memory.


Victor-C-Zhang (approved):

etwest:
Don't merge without merging the cache_aware gutter tree branch in GutterTree repo

PR #93, opened 2022-03-15
Switched to better make_supernode interface

etwest:
This allows us to specify a location for a Supernode to be placed for any for our calls to make_supernode

### DIFF
--- a/include/supernode.h
+++ b/include/supernode.h
@@ -62,18 +62,20 @@ private:
 
   Supernode(const Supernode& s);
 public:
-  static Supernode* makeSupernode(uint64_t n, long seed);
-  static Supernode* makeSupernode(uint64_t n, long seed, std::fstream &binary_in);
-
   /**
-   * Makes a supernode at the provided location (and does no error checking).
-   * @param loc     the memory location to put the supernode.
+   * Supernode construtors
    * @param n       the total number of nodes in the graph.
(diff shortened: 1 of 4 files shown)


include/supernode.h
```diff
@@ -62,18 +62,20 @@ class Supernode {
...
+   * Supernode construtors
    * @param n       the total number of nodes in the graph.
    * @param seed    the (fixed) seed value passed to each supernode.
-   * @return        a pointer to loc, the location of the supernode.
+   * @param loc     (Optional) the memory location to put the supernode.
+   * @return        a pointer to the newly created supernode object
    */
-  static Supernode* makeSupernode(void* loc, uint64_t n, uint64_t seed);
-  static Supernode* makeSupernode(const Supernode& s);
+  static Supernode* makeSupernode(uint64_t n, long seed, void *loc = malloc(bytes_size));
```
Victor-C-Zhang, line 72: could this ever be called when `bytes_size` is uninitialized?

etwest: Currently you have to call the configure function first. It's non-trivial to change that.

etwest: So yes

Victor-C-Zhang: not ideal, but i guess we can let it go for now. can you put up an issue for this?

Victor-C-Zhang (approved):
see comment reply

PR #97, opened 2022-03-24
feat: expose sketch specifics from supernode class

Victor-C-Zhang:
Distributing sketch queries needs to be able to get sketches from within a supernode. Right now, the necessary functions are private.

### DIFF
--- a/include/supernode.h
+++ b/include/supernode.h
@@ -51,6 +51,13 @@ private:
   Supernode(uint64_t n, uint64_t seed, std::istream &binary_in);
 
   Supernode(const Supernode& s);
+
+
+  // get the ith sketch in the sketch array
+  inline Sketch* get_sketch(size_t i) {
+    return reinterpret_cast<Sketch*>(sketch_buffer + i * sketch_size);
+  }
+  
 public:
   /**
    * Supernode construtors
@@ -85,9 +92,8 @@ public:
   // return the number of sketches held in this supernode
   int get_num_sktch() { return num_sketches; };
 
-    // get the ith sketch in the sketch array
-  inline Sketch* get_sketch(size_t i) {
-    return reinterpret_cast<Sketch*>(sketch_buffer + i * sketch_size);
+  inline bool out_of_queries() {
+    return idx == num_sketches;
   }
 
   // get the ith sketch in the sketch array
--- a/include/supernode.h
+++ b/include/supernode.h
@@ -50,16 +50,6 @@ private:
    */
   Supernode(uint64_t n, uint64_t seed, std::istream &binary_in);
 
-  // get the ith sketch in the sketch array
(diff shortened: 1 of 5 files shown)


include/supernode.h
```diff
@@ -84,10 +74,27 @@ class Supernode {
...
 
+  inline size_t get_sketch_size() {
+    return sketch_size;
+  }
+
+  // return the number of sketches held in this supernode
+  int get_num_sktch() { return num_sketches; };
+
+    // get the ith sketch in the sketch array
+  inline Sketch* get_sketch(size_t i) {
```
etwest: Is it possible to only make the const pointer public? This might require making Sketch write_binary const but that’s doable.

Victor-C-Zhang: I don't think that's how `const` member functions work. Afaik `const` versions of member functions are invoked only when the object is `const`. That means the `supernode` would need to be a `const` supernode, which clearly it cannot be since we are actively modifying it,

EDIT: Though now that I think about it, I could create a `get_curr_sketch()` function and just keep all native `get_sketch()` functions private.

Victor-C-Zhang: Why do you want this anyways?

etwest: I'm not saying the member function should be const but rather return a const pointer to a sketch object. The reason I want this is so that the caller of get_sketch can't muck with the internals of the supernode and make the data invalid for example.

etwest: Sketch const * is really what I want. Since we aren't returning a reference to the pointer we don't need to protect the pointer with const.

Victor-C-Zhang: Ok

Victor-C-Zhang: But this is impossible. We cannot have both a version of `get_sketch` that returns a const pointer and one that doesn't. Reason being that we need `Supernode` to be able to modify `Sketch` and we can't return 2 different things from the same method signature. Unless you really really want this compile-time safety measure, I don't really see how this would be beneficial for us.

etwest: Yeah... could also change the name of the public function. Maybe `read_sketch()` or something. I'll leave this up to you.

src/graph.cpp
```diff
@@ -23,7 +23,8 @@ Graph::Graph(node_id_t num_nodes): num_nodes(num_nodes) {
   supernodes = new Supernode*[num_nodes];
   parent = new node_id_t[num_nodes];
   size = new node_id_t[num_nodes];
-  seed = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+  seed = 69420;
```
etwest: TODO: remove this change before merging

Victor-C-Zhang: good catch

etwest (approved):
looks good

PR #98, opened 2022-03-28
Eager DSU Optimization

kennyzzhang:


src/graph.cpp
```diff
@@ -118,6 +122,23 @@ void Graph::generate_delta_node(node_id_t node_n, uint64_t node_seed, node_id_t
 void Graph::batch_update(node_id_t src, const std::vector<node_id_t> &edges, Supernode *delta_loc) {
   if (update_locked) throw UpdateLockedException();
 
+  for (auto it = edges.begin(); dsu_valid && it != edges.end(); ++it) {
```
etwest: Batch update is performed in parallel from many graph workers. Is this code thread safe?

etwest: Actually, it would probably be better to do the DSU update in graph::update. I know this post buffering thing was my idea but the main computation we want to avoid is flushing the buffers during queries. We need this in Graph::update if that'll happen.

include/graph.h
```diff
@@ -107,6 +113,26 @@ class Graph {
     gts->insert(edge);
     std::swap(edge.first, edge.second);
     gts->insert(edge);
+    if (dsu_valid) {
```
etwest, line 135: This should go above the calls to gts->insert()

kennyzzhang:
Running `./tests --gtest_filter=GraphTest.EagerDSUTest` never hangs, but running `./tests` sometimes hangs. A surefire way to get a hang is to run `./tests --gtest_filter=GraphTest.EagerDSUTest --gtest_repeat=10`, which might be helpful for debugging.

etwest (approved):
These are ready for main

PR #100, opened 2022-03-31
Multithreaded BinaryGraphStream

etwest:
Allows multiple threads to read from a BinaryGraphStream concurrently

### DIFF
--- a/include/binary_graph_stream.h
+++ b/include/binary_graph_stream.h
@@ -79,8 +79,10 @@ public:
     buf_size = _b - (_b % edge_size); // ensure buffer size is multiple of edge_size
 
     // read header from the input file
-    read(stream_fd, reinterpret_cast<char *>(&num_nodes), 4);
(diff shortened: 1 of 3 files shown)


DanielDeLayo (approved):

Victor-C-Zhang:
What do our benchmarks look like?
And why have we chosen to do this rather than have multiple readers reading from different points in the stream?

etwest:
> What do our benchmarks look like?
> And why have we chosen to do this rather than have multiple readers reading from different points in the stream?

It's a big improvement to performance especially when the file is pre-cached. We can do billions of updates per second from the in-memory file.

There are two reasons why this is a good approach. First off, we approximate having entirely independent stream readers by having them run independently upon 32 KiB buffers. Second, we can stop the stream at arbitrary points using this approach to perform queries. (Though we haven't implemented this yet)

PR #101, opened 2022-04-07
Fix Bug in MTBinaryStream

etwest:
This is a holdover bug from the previous BinaryGraphStream design. Previously we knew exactly when the stream would end (number of edge updates) so would stop asking for data. In this version we don't and use EOF to stop. However, we only check for EOF after inserting all the updates in a buffer. This led to more updates than expected being inserted.

This may explain the non-determinism in the results as the extra updates in the buffer that are erroneously inserted differ depending upon which thread is the last one to call read_data().


DanielDeLayo (approved):
looks good

PR #103, opened 2022-04-09
number of inserters pass through + virtual destructor

DanielDeLayo:
To initialize memory, we need to know the number of threads (and therefore number of buffers to the buffers).

This passes through that value.

Also, we were inheriting from a non-virtual destructor. This is undefined behavior.

When merging this, we should revert the change in the Distributed repo that points cmake to this branch.

### DIFF
--- a/include/graph.h
+++ b/include/graph.h
@@ -96,8 +96,8 @@ protected:
 
   static bool open_graph;
 public:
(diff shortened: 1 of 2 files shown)


etwest (approved):
Looks good

PR #105, opened 2022-04-24
Allow for programmatic graph configuration with file as fallback

etwest:
The current configuration scheme is annoying as the unit tests mess with the config file and it makes running experiments with consistent parameters difficult. These changes introduce a GraphConfiguration class that can be passed to the Graph constructor to determine the graph parameters programmatically.


include/graph_configuration.h
```diff
@@ -0,0 +1,93 @@
...
+    std::ifstream config_file(config_loc);
+    std::string line;
+    if (config_file.is_open()) {
+      while(getline(config_file, line)) {
+        if (line[0] == '#' || line[0] == '\n') continue;
+        if(line.substr(0, line.find('=')) == "buffering_system") {
+          std::string buf_str = line.substr(line.find('=') + 1);
+          if (buf_str == "tree") gutter_sys = GUTTERTREE;
+          else if (buf_str == "cachetree") gutter_sys = CACHETREE;
+          else if (buf_str != "standalone") {
```
DanielDeLayo: While this is correct, changing the default makes it incorrect.

src/graph.cpp
```diff
@@ -33,20 +34,33 @@ Graph::Graph(node_id_t num_nodes, int num_inserters): num_nodes(num_nodes) {
     supernodes[i] = Supernode::makeSupernode(num_nodes,seed);
     parent[i] = i;
   }
-  num_updates = 0; // REMOVE this later
```
DanielDeLayo, line 36: Why is this no longer temporary?

DanielDeLayo (approved):

include/graph_configuration.h
```diff
@@ -0,0 +1,93 @@
...
+  // Backup supernodes in memory or on disk when performing queries
+  bool backup_in_mem = true;
+
+  // The number of graph workers
+  size_t num_groups = 1;
+
+  // How many OMP threads each graph worker uses
+  size_t group_size = 1;
+
+  void read_configuration() {
```
Victor-C-Zhang, line 30: Instead of making the fields public and having a read option, can we just create two constructors; one where you provide all the values and one where you provide a file location?

include/graph_configuration.h
etwest: The default is STANDALONE, that seems like it should be correct to me.

src/graph.cpp
etwest: I added it a long time ago and thought it would just be a quick variable for testing but its actually really useful to have so we can just keep it.

include/graph_configuration.h
etwest: Oh, do you mean that changing the default gutter_sys would break this logic?

DanielDeLayo: Yes, exactly that.

DanielDeLayo (approved):

PR #107, opened 2022-05-06
test: graph multithreaded insert with standalone

Victor-C-Zhang:
Exactly as it sounds. There wasn't anything ensuring correctness of this feature so I wrote up a small sanity check


etwest (approved):
Looks good! Thanks

PR #108, opened 2022-05-06
feat: clearer pairing fn names

Victor-C-Zhang:
the current pairing function we use is not the non-self-edge pairing function we had used before. However, the original pairing function is still useful in some applications. This PR renames the current pairing function and restores the old pairing function


etwest (approved):
Looks good!

Victor-C-Zhang:
gonna pull from main before merging

Victor-C-Zhang:
actually jk it’s fine

PR #110, opened 2022-05-20
fix: sketch test configured vector sizes to match testing vector sizes

Victor-C-Zhang:
We were incorrectly conflating graph "size" (num_nodes) with sketch "size" (num_nodes*num_nodes) in sketch tests. Specifically, we were using graph size for the testing vector and sketch size for sketches. This means the sketches generated for testing were much larger than expected and so our failure rate was unintentionally too low because what we thought were "small" sizes were in fact not.


etwest (approved):
Good catch

PR #113, opened 2022-06-19
Fix for SketchTestLarge and updates to README

etwest:


Victor-C-Zhang:
can you explain the problem and fix for large sketch rest in more detail?

etwest:
> can you explain the problem and fix for large sketch rest in more detail?

So basically awhile ago I realized that when we're using the sketches for solving connectivity there's actually a tighter upper-bound on the number of non-zero elements in the vector than n^2, it's actually n^2 / 4. So we reduced the number of sketch guess levels correspondingly for a small performance gain and memory use reduction. However, when we sized the vectors down in the sketch test then we were filling up the vector beyond this lower bound. In expectation roughly half the items would be non-zero for the first test when vector size was 1,000,000 and number of updates was also 1,000,000

The fix I did was to simply reduce the number of updates by a factor 4 for the first test.

Victor-C-Zhang:
> > can you explain the problem and fix for large sketch rest in more detail?
> 
> So basically awhile ago I realized that when we're using the sketches for solving connectivity there's actually a tighter upper-bound on the number of non-zero elements in the vector than n^2, it's actually n^2 / 4. So we reduced the number of sketch guess levels correspondingly for a small performance gain and memory use reduction. However, when we sized the vectors down in the sketch test then we were filling up the vector beyond this lower bound. In expectation roughly half the items would be non-zero for the first test when vector size was 1,000,000 and number of updates was also 1,000,000
> 
> The fix I did was to simply reduce the number of updates by a factor 4 for the first test.

Is there any documentation in code or wiki that shows this upper bound?

etwest:
> > > can you explain the problem and fix for large sketch rest in more detail?
> > 
> > 
> > So basically awhile ago I realized that when we're using the sketches for solving connectivity there's actually a tighter upper-bound on the number of non-zero elements in the vector than n^2, it's actually n^2 / 4. So we reduced the number of sketch guess levels correspondingly for a small performance gain and memory use reduction. However, when we sized the vectors down in the sketch test then we were filling up the vector beyond this lower bound. In expectation roughly half the items would be non-zero for the first test when vector size was 1,000,000 and number of updates was also 1,000,000
> > The fix I did was to simply reduce the number of updates by a factor 4 for the first test.
> 
> Is there any documentation in code or wiki that shows this upper bound?

There's a comment in `sketch.h` that states it's true but there isn't a proof written up anywhere.

Victor-C-Zhang:
Maybe you should sketch a proof in the wiki for posterity. I can imagine we might want to extend it for hypergraphs or other situations when the analysis might change.

Victor-C-Zhang (approved):
see above comment

etwest:
[Proof Sketch](https://github.com/GraphStreamingProject/GraphZeppelin/wiki/Proof-of-Tighter-Upper-Bound-on-Non-Zero-Elements-for-CC)

PR #114, opened 2022-07-10
Add functionality for queries during the stream when reading from stream with multiple threads

etwest:
Two types of query:
- On demand query: Execute the query immediately after all threads finish processing their current blocks of data
- Pre-registered query: Execute a query after a specific update in the stream. Must give advance notice of this query (if a thread is processing the block this query index is within than the request to pre-register will fail.


include/binary_graph_stream.h
```diff
@@ -76,6 +76,7 @@ class BinaryGraphStream_MT {
       throw BadStreamException();
     }
 
+    // set the buffer size to be a multiple of an edge size
     buf_size = _b - (_b % edge_size); // ensure buffer size is multiple of edge_size
```
Victor-C-Zhang: same comment

include/binary_graph_stream.h
```diff
@@ -84,31 +85,80 @@ class BinaryGraphStream_MT {
...
+  // This allows queries to be performed in the stream arbitrary at 32 KiB granularity
+  // without advance notice
+  void on_demand_query() { query_block = true; }
+
+  // call this function to tell stream its okay to keep going
+  // call once per query performed regardless if registered query or on-demand query
+  void post_query_resume() { query_block = false; query_index = -1; }
+
+  /*
+   * Call this function to register a query in advance to avoid contraint on 32 KiB boundary
```
Victor-C-Zhang: *constraint

include/binary_graph_stream.h
```diff
@@ -84,31 +85,80 @@ class BinaryGraphStream_MT {
...
+   * Only one query may be registered at a time and query index must within a 32 KiB boundary not
+   * yet touched by the MT_StreamReader threads.
+   * To register first query, call before processing any updates from the stream
+   * When registering second (or later) query, call register_query after post_query_resume but 
+   * before calls to get_edge() to ensure the registration and query succeed
+   * @param query_idx   the stream update index directly after which the query will be performed
+   * @return            true if the query is successfully registered and false if not
+  */
+  bool register_query(uint64_t query_idx) {
+    uint64_t byte_index = 12 + query_idx * edge_size;
```
Victor-C-Zhang: turn this magic number into a var

include/binary_graph_stream.h
```diff
@@ -84,31 +85,80 @@ class BinaryGraphStream_MT {
...
+    if (byte_index <= stream_off) return false;
+    else query_index = byte_index;
+    return true;
+  }
+
   inline void stream_reset() {stream_off = 12;}
```
Victor-C-Zhang: same with this instance (and others if there are any)

include/binary_graph_stream.h
```diff
@@ -84,31 +85,80 @@ class BinaryGraphStream_MT {
...
   inline uint32_t read_data(char *buf) {
+    // we are blocking on a query or the stream is done so don't fetch_add or read
+    if (query_block || stream_off >= end_of_file || stream_off >= query_index) return 0;
+
+    // multiple threads may execute this line of code at once. This can cause edge cases
     uint64_t read_off = stream_off.fetch_add(buf_size, std::memory_order_relaxed);
+
+    // we catch these edge cases using the two below checks
+    if (read_off >= query_index) {
+      stream_off = query_index.load();
```
DanielDeLayo, line 154: I'm concerned about a race condition here when unpausing.

Consider two threads that are assigned a read past the query as it's finishing. The query is atomically removed/changed/etc, but only one of the two threads read the query offset before it was changed. One sets the stream offset back, while the other proceeds with unaccounted for work. Then, more threads come in and eventually replay the work of the second thread, cancelling it out.

It's an unlikely case (especially if there's locking outside this function or queryblock is true), but still possible.
It's one of the issues caused by using two atomics together. 

I'm not sure there's an obvious fix without locks.

DanielDeLayo (dismissed):

include/types.h
```diff
@@ -6,8 +6,12 @@ typedef uint64_t col_hash_t;
...
+// Is a stream update an insertion or a deletion
+// NXT_QUERY: special type to indicate getting more data would cross a query boundary
+// END_OF_FILE: special type to indicate that there is no more data in the stream
 enum UpdateType {
   INSERT = 0,
   DELETE = 1,
-  END_OF_FILE = 2
-}; // special type to indicate that there is no more data in the stream
+  NXT_QUERY,
+  END_OF_FILE
```
Victor-C-Zhang: nit: give these explicit numbers too

test/graph_test.cpp
```diff
@@ -341,3 +342,117 @@ TEST(GraphTest, MultipleInsertThreads) {
...
+    int num_query_ready = 0;
+    std::condition_variable q_ready_cond;
+    std::condition_variable q_done_cond;
+    std::mutex q_lock;
+
+    // prepare evenly spaced queries
+    int num_queries = 10;
+    int upd_per_query = num_edges / num_queries;
+    int query_idx = upd_per_query;
+    ASSERT_EQ(stream.register_query(query_idx), true); // register first query
```
Victor-C-Zhang: ASSERT_TRUE

test/graph_test.cpp
```diff
@@ -341,3 +342,117 @@ TEST(GraphTest, MultipleInsertThreads) {
...
+
+TEST(GraphTest, MTStreamWithMultipleQueries) {
+  for(int i = 1; i <= 10; i++) {
+    GraphConfiguration config;
+    config.gutter_sys = STANDALONE;
+
+    const std::string fname = __FILE__;
+    size_t pos = fname.find_last_of("\\/");
+    const std::string curr_dir = (std::string::npos == pos) ? "" : fname.substr(0, pos);
+    BinaryGraphStream_MT stream(curr_dir + "/res/multiples_graph_1024.data", 256);
```
Victor-C-Zhang: consider doing this test with a random graph or graph with deletions

Victor-C-Zhang (approved):
some small things

test/graph_test.cpp
etwest: We currently have no ability to define random binary streams. However, it wouldn't be hard to add some deletions to the .data graph stream

etwest: I'll add that as an issue

EDIT: actually ill just fix it now

include/binary_graph_stream.h
etwest: This is true. However, an important requirement to perform a query is that all threads have stopped doing updates (therefore no more calls to read_data()). Assuming that the user of the stream is handling this correctly all we need to worry about here is when a bunch of threads call fetch_add at once when stream_off < query_index. 

We can safely set back to query_index because that's where the stream should stop pre-query. No threads should be calling read_data() during the query so there shouldn't be a race on query_index being changed in a call to register_query() or post_query_resume(). It's the responsibility of the thread performing the query to perform the necessary updates to the query_index before unblocking the other threads.

Although, I'm not super happy with this design as it is very reliant on being used correctly. Do you have any ideas for de-facto safer design or a way to force users to interact with the stream in a safe manner?

DanielDeLayo: From what you've told me, it sounds like the system is almost workable. We'll need to add documentation/comments/preconditions explaining how to use this function properly (assuming I haven't missed it), as well as a readers-writer lock.

Specifically, a "Read-preferring RW lock."

https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock

There may be standard implementations, but here's one I'm thinking of:

Each thread stores a 'I'm within the critical zone' atomic. Assuming no writers, this atomic can safely be stored in L1 with no contention. Then, when a query occurs, whichever thread is responsible for it (I assume this is a single thread here) iterates through all the relevant atomics, locking them and preventing them from passing. After the 'Query' thread locks all other threads, we know that no other threads can be within the critical zone, regardless of outside implementation.

Let me know what you think.

etwest: I think that would work but it seems like an unnecessary level of synchronization. Basically the caller already needs to ensure that all threads stop processing updates in order for the query to be valid. Perhaps there is a way to hit both these 'birds' with one 'stone'.

For now would it suffice for me to add some comments explaining this in the code and perhaps open an issue to come up with a better solution later?

DanielDeLayo: That works for me

PR #124, opened 2023-02-15
Faster fast with branch prediction

etwest:
Gotta go faster


DanielDeLayo (approved):
That's a lot of fast. Do we want to have the PHT instead of the Standalone Gutters in config?

etwest:
> That's a lot of fast. Do we want to have the PHT instead of the Standalone Gutters in config?

No that was a mistake

PR #126, opened 2023-03-21
Supernode: range merge function

etwest:
Ability to merge together only some of the sketches in two Supernodes.


DanielDeLayo (approved):
The range merge worries me a little bit, but I think it's probably fine. We trust user-given serialization streams, so trusting a range-merge call isn't worse.

PR #131, opened 2023-06-24
Verifier changes

etwest:
The verifier is a little overpowered and as a consequence there are a bunch of race conditions if edges are verified by multiple threads. This pull request reduces the power of the verifier while ensuring it still catches bugs. The verifier now checks the following:
 - Edge correctness, each edge found in queries is validated as being within the graph
 - CC correctness, the connected components returned by the query are verified to be correct.

The verifiers no longer check that no redundant edges are passed to `verify_edge` this means the verifier has to run a dsu which is a little overpowered and introduces race conditions. We already have tests for the DSU so it's probably fine.

### DIFF
--- a/include/test/file_graph_verifier.h
+++ b/include/test/file_graph_verifier.h
@@ -11,14 +11,11 @@
  */
 class FileGraphVerifier : public GraphVerifier {
   std::vector<std::set<node_id_t>> kruskal_ref;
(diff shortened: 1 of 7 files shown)


DanielDeLayo (approved):
Reducing the power of the verifier is probably fine.

PR #132, opened 2023-07-02
Bug fix for Multithreaded DSU

etwest:
This pull request fixes an edge case bug in the multithreaded DSU. Specifically the error occurs under the following conditions.

1. Two threads find the same edge and attempt to perform a DSU merge
2. The endpoints of the edge (x, y) have the same component size, i.e. size[x] == size[y]
3. By chance, thread 1 makes y the parent of x and simultaneously thread 2 makes x the parent of y.

This will not cause the CAS to fail because thread 1 is operating on parent[x] while thread 2 is operating upon parent[y]. To solve this issue we make the ordering of edges deterministic and fixed. Thus, both threads orient the edge the same way if the sizes are identical. To "randomize" the roots, we switch whether the smaller or larger of x/y is first based upon if their sum is even/odd.

### DIFF
--- a/include/dsu.h
+++ b/include/dsu.h
@@ -25,11 +25,15 @@ class DisjointSetUnion {
   // if sum odd, larger first
   inline void order_edge(T& a, T& b) {
     if (size[a] < size[b]) std::swap(a, b);
-    if (size[a] == size[b]) {
-      if ((a + b) % 2 == 0 && a > b)
-        std::swap(a, b);
-      else if ((a + b) % 2 == 1 && a < b)
-        std::swap(a, b);
+    else if (size[a] == size[b]) {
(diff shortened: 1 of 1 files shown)


Victor-C-Zhang:
why is this an issue? aren’t we always merging into the lexicographically lower parent in cases of ties?

etwest:
So the way that the code has worked so far is that if the sizes are the same then whichever endpoint is first (the src) then that is made the parent. We found it was better to "randomize" which endpoint was first so that the roots aren't clustered. This helps performance but also causes the issue I'm talking about in this pull request. Feels like this should just be solved on the DSU side rather than the caller having to worry about it.

DanielDeLayo (approved):

PR #133, opened 2023-07-11
Support for allocating more sketches than normal

etwest:
We need to, for a variety of potential applications, be able to allocate more than just O(log n) sketches. This pull request gives that option to users through the `graph_configuration`.

### DIFF
--- a/include/graph_configuration.h
+++ b/include/graph_configuration.h
@@ -30,6 +30,12 @@ private:
   // How many OMP threads each graph worker uses
   size_t _group_size = 1;
 
+  // Option to create more sketches than for standard connected components
+  // Ex factor of 1, double the sketches
+  //    factor of 0.5, 1.5 times the sketches
+  //    factor of 0, normal quantity of sketches
+  double _adtl_skts_factor = 0;
+
   // Configuration for the guttering system
   GutteringConfiguration _gutter_conf;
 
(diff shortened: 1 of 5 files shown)


include/graph_configuration.h
```diff
@@ -30,6 +30,12 @@ class GraphConfiguration {
   // How many OMP threads each graph worker uses
   size_t _group_size = 1;
 
+  // Option to create more sketches than for standard connected components
+  // Ex factor of 1, double the sketches
+  //    factor of 0.5, 1.5 times the sketches
+  //    factor of 0, normal quantity of sketches
+  double _adtl_skts_factor = 0;
```
Victor-C-Zhang: my engineer brain dislikes this very much. would be simpler for it to be just multiplicative and name it accordingly

Victor-C-Zhang (dismissed):

Victor-C-Zhang:
pls also document the changes you made to memory profiling/etc

etwest:
> pls also document the changes you made to memory profiling/etc

What changes are you referring to exactly?

include/graph_configuration.h
etwest: Hmm okay I suppose that works at well.

etwest: Done

Gillgamesh (approved):

PR #135, opened 2023-10-18
Refactor Graph class -> GraphSketchDriver + CCSketchAlg

### LINKED ISSUE #96: Supernode and Sketch Configure
These functions have to be called before make_supernode or make_sketch respectively. It would be good to redesign the code so that these functions are not necessary and the parameters they set are automatically determined as required.

### LINKED ISSUE #115: Setup for queries during multi-threaded stream is awkward
Currently the setup in binary_graph_stream.h is awkward. It requires the caller to be very careful about ensuring that threads are correctly paused before doing the query. 

It would be much nicer to have a setup where we just call a query function and all this is handled behind the scenes.

etwest:
This pull request refactors the key classes of this repository with the goal of increasing the flexibility of our code. We want to allow the possibly for writing code for other sketching algorithms on top of this repository. Specifically, we disentangled the setup for reading and processing graph streams from the code for running the connected components sketch algorithm. We also got rid of the Supernode class and refactored the sketch class to be more usable.

Also some other small stuff, see below.
- [x] Config for driver and cc algorithm seperately
- [x] Extract spanning forest edges option for CCSketchAlg
- [x] Sketch constructor using vector length rather than number of vertices. Add static vec_len_from_verts().
- [x] Graph destructor
- [x] Raw bucket delta update
- [x] Exaustive sample
- [x] Investigate query performance
- [x] More testing
- [x] Rename graph_test.cpp to cc_alg_test.cpp 
- [x] See comments below


include/sketch.h
```diff
@@ -0,0 +1,144 @@
...
+ * Sketch for graph processing, either CubeSketch or CameoSketch.
+ * Sub-linear representation of a vector.
+ */
+class Sketch {
+ private:
+  const uint64_t seed;     // seed for random number generators
+  size_t num_samples;      // number of samples we can perform
+  size_t cols_per_sample;  // number of columns to use on each sample
+  size_t num_columns;      // Total number of columns. (product of above 2)
+  size_t num_guesses;      // number of buckets per column
```
DanielDeLayo: rename to buckets_per_column or something

include/graph_configuration.h
```diff
@@ -37,26 +34,28 @@ class GraphConfiguration {
   // Configuration for the guttering system
   GutteringConfiguration _gutter_conf;
 
-  friend class Graph;
+  friend class CCSketchAlg;
```
DanielDeLayo: TODO: This should be separated into 2 configs?

include/sketch.h
```diff
@@ -0,0 +1,144 @@
...
+  void serialize(std::ostream& binary_out) const;
+
+  inline void reset_sample_state() {
+    sample_idx = 0;
+  }
+
+  // return the size of the sketching datastructure in bytes (just the buckets, not the metadata)
+  inline size_t sketch_bytes() const { return num_buckets * sizeof(Bucket); }
+
+  inline const char* get_bucket_memory_buffer() const { return (char*)buckets; }
```
DanielDeLayo: get_readonly_buckets())

include/sketch.h
```diff
@@ -0,0 +1,144 @@
...
+  inline uint64_t get_seed() const { return seed; }
+  inline size_t column_seed(size_t column_idx) const { return seed + column_idx * 5; }
+  inline size_t checksum_seed() const { return seed; }
+  inline size_t get_columns() const { return num_columns; }
+
+  // max number of non-zeroes in vector is n/2*n/2=n^2/4
+  static size_t guess_gen(size_t n) { return 2 * log2(n) - 2; }
```
DanielDeLayo: calc_buckets_per_col()

include/sketch.h
```diff
@@ -0,0 +1,144 @@
...
+  static size_t samples_gen(size_t n) { return ceil(log2(n) / num_samples_div); }
```
DanielDeLayo: calc_max_samples()?

include/worker_thread_group.h
```diff
@@ -0,0 +1,168 @@
...
+          break;
+        }
+      }
+      lk.unlock();
+
+      if (all_paused) break;  // all workers are done so exit
+    }
+  }
+  void resume_workers() {
+    // unpause the WorkerThreads
```
DanielDeLayo, line 163: gts->set_non_block(false);

src/cc_sketch_alg.cpp
```diff
@@ -36,134 +21,138 @@ Graph::Graph(node_id_t num_nodes, GraphConfiguration config, int num_inserters)
...
   spanning_forest_mtx = new std::mutex[num_nodes];
   dsu_valid = true;
+  shared_dsu_valid = true;
+  num_updates_per_batch = delta_sketches[0]->sketch_bytes() / sizeof(node_id_t);
+  num_updates_per_batch *= config._batch_factor;
   std::cout << config << std::endl;  // print the graph configuration
 }
 
-Graph::Graph(const std::string &input_file, GraphConfiguration config, int num_inserters)
+CCSketchAlg::CCSketchAlg(std::string input_file, GraphConfiguration config)
```
DanielDeLayo: const reference input_file

src/cc_sketch_alg.cpp
```diff
@@ -36,134 +21,138 @@ Graph::Graph(node_id_t num_nodes, GraphConfiguration config, int num_inserters)
...
-  if (open_graph) throw MultipleGraphsException();
-
-  vec_t sketch_num_columns;
   double sketches_factor;
   auto binary_in = std::fstream(input_file, std::ios::in | std::ios::binary);
   binary_in.read((char *)&seed, sizeof(seed));
   binary_in.read((char *)&num_nodes, sizeof(num_nodes));
-  binary_in.read((char *)&sketch_num_columns, sizeof(sketch_num_columns));
   binary_in.read((char *)&sketches_factor, sizeof(sketches_factor));
-  Supernode::configure(num_nodes, sketch_num_columns, sketches_factor);
```
DanielDeLayo: Whiteboard graph serialization

src/cc_sketch_alg.cpp
```diff
@@ -36,134 +21,138 @@ Graph::Graph(node_id_t num_nodes, GraphConfiguration config, int num_inserters)
...
   representatives = new std::set<node_id_t>();
-  supernodes = new Supernode *[num_nodes];
+  sketches = new Sketch *[num_nodes];
   parent = new std::remove_reference<decltype(*parent)>::type[num_nodes];
   size = new node_id_t[num_nodes];
   std::fill(size, size + num_nodes, 1);
   for (node_id_t i = 0; i < num_nodes; ++i) {
     representatives->insert(i);
-    supernodes[i] = Supernode::makeSupernode(num_nodes, seed, binary_in);
+    sketches[i] = new Sketch(num_nodes, seed, binary_in, FULL);
```
DanielDeLayo: get rid of FULL

src/cc_sketch_alg.cpp
```diff
@@ -36,134 +21,138 @@ Graph::Graph(node_id_t num_nodes, GraphConfiguration config, int num_inserters)
...
+          spanning_forest[src].insert(dst);
+          break;
+        }
+      }
     }
   }
-  Supernode::delta_supernode(node_n, node_seed, updates, delta_loc);
+#else
+  (void)upd;
+  unlikely_if(dsu_valid) {
```
DanielDeLayo: maybe add a comment about cache coherency here?

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+    : seed(seed) {
+  num_samples = _samples == 0 ? samples_gen(n) : _samples;
+  cols_per_sample = _cols == 0 ? default_cols_per_sample : _cols;
+  num_columns = num_samples * cols_per_sample;
+  num_guesses = guess_gen(n);
+  num_buckets = num_columns * num_guesses + 1; // plus 1 for deterministic bucket
+  buckets = new Bucket[num_buckets];
+
+  // Read the serialized Sketch contents
+  binary_in.read((char *)buckets, num_buckets * sizeof(Bucket));
```
DanielDeLayo: consider using sketch size method here and below

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+}
+#endif
+
+void Sketch::zero_contents() {
+  for (size_t i = 0; i < num_buckets; i++) {
+    buckets[i].alpha = 0;
+    buckets[i].gamma = 0;
+  }
```
DanielDeLayo, line 90: call reset samples?

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+      unsigned bucket_id = (i + first_column) * num_guesses + j;
+      if (Bucket_Boruvka::is_good(buckets[bucket_id], checksum_seed()))
+        return {buckets[bucket_id].alpha, GOOD};
+    }
+  }
+  return {0, FAIL};
+}
+
+std::pair<std::unordered_set<vec_t>, SampleSketchRet> Sketch::exhaustive_sample() {
+  // TODO!
```
DanielDeLayo: TODO!

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+  exit(EXIT_FAILURE);
+}
+
+void Sketch::merge(Sketch &other) {
+  if (other.buckets[num_buckets-1].alpha == 0 && other.buckets[num_buckets-1].gamma == 0) {
+    // other sketch is empty so just return
+    return;
+  }
```
DanielDeLayo, line 133: consider deletion or unlikely_if

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  // perform the merge
+  for (size_t i = 0; i < num_buckets; ++i) {
+    buckets[i].alpha ^= other.buckets[i].alpha;
+    buckets[i].gamma ^= other.buckets[i].gamma;
+  }
+}
+
+void Sketch::merge_raw_bucket_buffer(vec_t *buckets, size_t start_sample, size_t num_samples) {
+  // TODO!
```
DanielDeLayo: TODO!1!

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+  exit(EXIT_FAILURE);
+}
+
+void Sketch::serialize(std::ostream &binary_out) const {
+  binary_out.write((char*) buckets, num_buckets * sizeof(Bucket));
```
DanielDeLayo: use size function

src/sketch.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+bool operator==(const Sketch &sketch1, const Sketch &sketch2) {
+  if (sketch1.num_buckets != sketch2.num_buckets || sketch1.seed != sketch2.seed) {
+    std::cout << "sketch1 = " << sketch1 << std::endl;
+    std::cout << "sketch2 = " << sketch2 << std::endl;
+    return false;
+  }
```
DanielDeLayo: delete debug output
consider assert?

test/graph_test.cpp
```diff
@@ -32,79 +32,51 @@ TEST_P(GraphTest, SmallGraphConnectivity) {
...
-    g.should_fail_CC();
+    cc_alg.should_fail_CC();
+
+    // flush to make sure copy sketches is consistent with graph sketches
+    GraphSketchDriver<CCSketchAlg> driver(&cc_alg, &stream, config);
+    driver.process_stream_until(END_OF_STREAM);
+
+    // manually ensure we have flushed
+    driver.gts->force_flush();
+    driver.worker_threads->flush_workers();
```
DanielDeLayo: call prep_query instead

DanielDeLayo (changes requested):

DanielDeLayo:
We need to change the sketch interface. Let us input the vector length, instead of the square root of the vector length.

test/graph_test.cpp
etwest: prep_query() doesn't do what we want because of the dsu optimization. So we need to do it manually.

DanielDeLayo (dismissed):
The .yml file is much more readable now!

include/cc_alg_configuration.h
```diff
@@ -0,0 +1,40 @@
...
+
+  // getters
+  std::string get_disk_dir() { return _disk_dir; }
+  double get_sketch_factor() { return _sketches_factor; }
+  double get_batch_factor() { return _batch_factor; }
+
+  friend std::ostream& operator<< (std::ostream &out, const CCAlgConfiguration &conf);
+
+  // no use of equal operator
+  CCAlgConfiguration& operator=(const CCAlgConfiguration &) = delete;
```
DanielDeLayo: Why are we against move and copy assignment operators? Is it the ambiguity?

(My reading of the spec tells me disabling copy-assign also disabled move-assign)

include/cc_sketch_alg.h
```diff
@@ -0,0 +1,216 @@
+#pragma once
+
+#include <atomic>  // REMOVE LATER
```
DanielDeLayo: // REMOVE LATER

src/cc_sketch_alg.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include <map>
+#include <random>
+
+CCSketchAlg::CCSketchAlg(node_id_t num_nodes, CCAlgConfiguration config)
+    : num_nodes(num_nodes), dsu(num_nodes), config(config) {
+  representatives = new std::set<node_id_t>();
+  sketches = new Sketch *[num_nodes];
+  seed = std::chrono::duration_cast<std::chrono::microseconds>(
+             std::chrono::high_resolution_clock::now().time_since_epoch())
+             .count();
```
DanielDeLayo: We should require the seed be given as input. While microseconds might be enough when using multiple CCSketchAlgs to avoid most duplicate seeds, I'd rather give the user control of the randomness generation.

src/sketch.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+Sketch::Sketch(vec_t vector_len, uint64_t seed, std::istream &binary_in, size_t _samples,
+               size_t _cols)
+    : seed(seed) {
+  num_samples = _samples;
+  cols_per_sample = _cols;
+  num_columns = num_samples * cols_per_sample;
+  bkt_per_col = calc_bkt_per_col(vector_len);
+  num_buckets = num_columns * bkt_per_col + 1; // plus 1 for deterministic bucket
+  buckets = new Bucket[num_buckets];
```
DanielDeLayo, line 31: I'm tempted to move the deterministic bucket to Sketch, so you can access it without a pointer lookup.
The main downside of this is getting a list of buckets would be more difficult.

tools/benchmark/graphcc_bench.cpp
```diff
@@ -50,12 +50,19 @@ static void BM_FileIngest(benchmark::State& state) {
...
-    while (m--) {
-      benchmark::DoNotOptimize(upd = stream.get_edge());
+    BinaryFileStream stream("/mnt/ssd2/binary_streams/kron_15_stream_binary");
+
+    bool reading = true;
+    while (reading) {
+      GraphStreamUpdate upds[state.range(0)];
+      size_t num_updates = stream->get_update_buffer(upds, state.range(0));
+      for (size_t i = 0; i < num_updates; i++) {
+        GraphStreamUpdate &upd = upds[i];
```
DanielDeLayo, line 60: DoNotOptimise removed-- making sure this is intentional

tools/test_correctness.cpp
```diff
@@ -0,0 +1,90 @@
...
+  std::vector<node_id_t> vertices;
+  for (size_t i = 0; i < num_vertices; i++) {
+    vertices.push_back(i);
+  }
+
+  size_t t = std::chrono::steady_clock::now().time_since_epoch().count();
+  std::mt19937_64 gen(t);
+
+  for (size_t g = 0; g < num_graphs; g++) {
+    size_t edge_seed = gen();
```
DanielDeLayo, line 38: there's no parallelism here-- can we not use gen for all of our randomness? Why do we reseed with a random number every iteration?

Randomly generating seeds (as well as sequential seeds) are both a bit messy. (birthday paradox and risk of correlation)

I don't think it matters here too much, but every now and then I enter into the raddit hole of using RNG to seed RNG and exit without a clear indication of what's best. This post claims to know more than I do.

https://stats.stackexchange.com/questions/233061/best-way-to-seed-n-independent-random-number-generators-from-1-value

DanielDeLayo (changes requested):

src/cc_sketch_alg.cpp
etwest: Agreed!

tools/benchmark/graphcc_bench.cpp
etwest: yes intentional

src/sketch.cpp
etwest: Nah

DanielDeLayo (approved):

PR #136, opened 2023-11-09
Added functions to perform a decomposed update

rebmaq:
Added 3 different functions that perform different parts of a sketch update to avoid recalculating hashes when performing sketch updates, in cases where you want to apply the same update to sketches with the same randomness,

### DIFF
--- a/include/sketch.h
+++ b/include/sketch.h
@@ -155,9 +155,11 @@ class Sketch {
     sample_idx = 0;
   }
 
+  #ifndef L0_SAMPLING
   vec_hash_t get_checksum(const vec_t update_idx);
   std::vector<size_t> const get_bucket_ids(const vec_t update_idx);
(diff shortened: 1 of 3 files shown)


DanielDeLayo:
1. We should rename SketchDelta to SketchHashes to be unambiguous.
2. There's two main paths forward in this implementation. One is to abandon buffering, allocate this SketchHashes once per thread, and use that. The other is two have some object construct and manipulate appropriately sized chunks of memory.

etwest (changes requested):
What Daniel said. Also I'd suggest we just store the depth/hash function results instead of worrying about transforming that to bucket_ids.

DanielDeLayo:
I think we decided that, as implemented, it didn't help performance.
This pull request has become stale-- I'm going to close it for now.
If we want to re-pursue this, we can reopen.

PR #139, opened 2024-01-22
statistical testing

DanielDeLayo:
This tests the probability that a column will succeed given z non-zeros. 
There is independence for a given z; we repeat for statistical significance.
There is likely not independence when varying z; for performance, the data point at z is based on the same state for the data point at z-1.


etwest:
Another pull request deletes the statistical testing directory. I think this should just live in the tools directory.

DanielDeLayo:
What's the scope of the new repository? Should this live there instead?

etwest:
> What's the scope of the new repository? Should this live there instead?

The new repository is just for stream generation and parsing. I didn't move the other statistical tests there. I just deleted them.

etwest (changes requested):
Looks overall good. Just add a little comment block at the top of both of these files to say what they do and we'll merge this in.

etwest (approved):
Yaaaaaaayyy....

PR #145, opened 2024-02-21
Make sketches_factor actually do something and fix exception race-case

etwest:
Main branch has problems:
- Currently we have the sketches factor option in the configuration but it doesn't affect anything... oops.
- When our threads throw an exception this can cause a race case resulting in memory errors instead of exiting gracefully. 
- log2(const expr) is not always a const expr with all compilers.
- The pre_insert function builds a representation of the entire graph on an insert only stream. yikes. Instead we want to just maintain the spanning forest edges.

This pull request fixes all of those problems.

### DIFF
--- a/tools/process_stream.cpp
+++ b/tools/process_stream.cpp
@@ -85,7 +85,7 @@ int main(int argc, char **argv) {
   std::cout << std::endl;
 
   auto driver_config = DriverConfiguration().gutter_sys(CACHETREE).worker_threads(num_threads);
-  auto cc_config = CCAlgConfiguration().batch_factor(1);
+  auto cc_config = CCAlgConfiguration().batch_factor(1).sketches_factor(0.5);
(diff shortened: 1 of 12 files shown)


tools/process_stream.cpp
```diff
@@ -85,7 +85,7 @@ int main(int argc, char **argv) {
   std::cout << std::endl;
 
   auto driver_config = DriverConfiguration().gutter_sys(CACHETREE).worker_threads(num_threads);
-  auto cc_config = CCAlgConfiguration().batch_factor(1);
+  auto cc_config = CCAlgConfiguration().batch_factor(1).sketches_factor(0.5);
```
DanielDeLayo: I'm not entirely convinced we want this as the default

DanielDeLayo (changes requested):

DanielDeLayo (approved):

PR #146, opened 2024-03-05
Better verifier

etwest:
* `GraphSketchDriver` builds a verifier and supplies it to the algorithm.
*  `GraphSketchDriver` can check its verifier against a "ground truth" verifier. For example, from a cumulative file.
*  `GraphSketchDriver` now passes an integer query code to the algorithm's `has_cached_query(int)` function so that the algorithm can determine on a query type basis whether it has an answer cached. For example, we can currently cache a single spanning forest but not k spanning forests.
* File and Matrix verifiers combined into a single verifier class.
* `Sketch::calc_cc_samples` always returns a number of samples `>= 18` to avoid errors for small graphs in our unit tests
* Added spanning forest verification functions to the verifier.
* Added unit testing for the verifier, for spanning forest generation, and for insert only streams.


include/sketch.h
```diff
@@ -72,7 +72,7 @@ class Sketch {
    * @return               The number of samples
    */
   static size_t calc_cc_samples(node_id_t num_vertices, double f) {
-    return ceil(f * log2(num_vertices) / num_samples_div);
+    return std::max(size_t(18), (size_t) ceil(f * log2(num_vertices) / num_samples_div));
```
DanielDeLayo, line 75: I want to make sure I fully understand this change before I approve it

etwest: This is enforcing that every connectivity sketch contains at least 18 columns. I made this change because at the small sizes our unit tests are running our failure rate is a bit too high. As the number of vertices increases the failure rate should decrease so this shouldn't be hiding any errors.

DanielDeLayo (approved):

PR #148, opened 2024-03-27
Eager Empty Bucket Checking

Gillgamesh:
Implements the following changes:

1. The function `effective_size()` tracks the number of rows in a sketch column that are at or below a nonzero bucket. That is, it tracks (index of deepest nonempty bucket) + 1.
2. There is an optional flag to enable the tracking of a `nonempty_buckets` bit array. This has the effect of making `effective_size()` take a constant number of instructions by `clz`. The flag is updating on calls to the three merge functions and to `update()`.
3. `sample()` and `exhaustive_sample()` have been updated to only check a small constant number of buckets below `effective_size()` for each column. This is based off a number of theoretical observations about the concentration of good buckets near the "best" bucket, on both the right and left. This should speed up calls to them by a 2-4x factor without sacrafices to the success probability.
4. `merge()` and `range_merge()` now use `effective_size()` to not bother merging in buckets that are definitely zero. 
5. Similarly, `serialize()` and the deserialize constructor also use `effective_size()` to shrink the space needed for storing and sending sketches.

The performance testing needs to be done thorouhgly, but this should slightly speed up merging, signifcantly speed up querying, and slightly slow down updating.


etwest:
Since this pull request is modifying the `Sketch` class which is essential to correctness and performance I think we should be extremely thorough in our review. This means we may end up making a lot of critiques/suggestions. Just want to set that expectation before we start.

include/bucket.h
```diff
@@ -49,6 +49,10 @@ namespace Bucket_Boruvka {
    */
   inline static void update(Bucket& bucket, const vec_t update_idx,
                             const vec_hash_t update_hash);
+  inline static bool is_empty(const Bucket &bucket) {
```
etwest: Implementation should go below like with the the other functions.

include/bucket.h
```diff
@@ -49,6 +49,10 @@ namespace Bucket_Boruvka {
...
+    // return bucket.alpha == 0 && bucket.gamma == 0;
```
etwest: Drop comment

include/sketch.h
```diff
@@ -125,6 +125,18 @@ class Sketch {
 
   std::mutex mutex; // lock the sketch for applying updates in multithreaded processing
 
+
+  /**
```
etwest, line 192: Line up `*` and provide a little more documentation. Specify that this operates per column and that all non-empty buckets are above this cutoff.

etwest: Better name?

src/sketch.cpp
```diff
@@ -18,6 +27,14 @@ Sketch::Sketch(vec_t vector_len, uint64_t seed, size_t _samples, size_t _cols) :
     buckets[i].alpha = 0;
     buckets[i].gamma = 0;
   }
+
+  #ifdef EAGER_BUCKET_CHECK
```
etwest: We want to have two forms of serialization:
  1. Direct "no-copy" serialization where we pull the bucket array data out directly.
  2. Compressed serialization which leverages the known column sizes to be more data size efficient.

To pull this off we need the buckets and flags to be stored contiguously. Suggest allocating a few extra buckets that are then used for the flags.

NOTE: We will all have to think about the right way to "receive" these serialized forms on the other side.

src/sketch.cpp
```diff
@@ -43,9 +71,21 @@ Sketch::Sketch(const Sketch &s) : seed(s.seed) {
   buckets = new Bucket[num_buckets];
 
   std::memcpy(buckets, s.buckets, bucket_array_bytes());
+
+  #ifdef EAGER_BUCKET_CHECK
```
etwest, line 133: Once we make flags contiguous we can skip this extra `memcpy` so long as we update `bucket_array_bytes()` to include the flags.

src/sketch.cpp
```diff
@@ -78,6 +118,13 @@ void Sketch::update(const vec_t update_idx) {
     size_t bucket_id = i * bkt_per_col + depth;
     likely_if(depth < bkt_per_col) {
       Bucket_Boruvka::update(buckets[bucket_id], update_idx, checksum);
+      #ifdef EAGER_BUCKET_CHECK
+      likely_if(!Bucket_Boruvka::is_empty(buckets[bucket_id])) {
```
etwest: `unlikely_if(is_empty)`

etwest: Investigate the performance of the following:
```
unlikely_if(is_empty)
  set_bit()
update()
unlikely_if(is_empty)
  clear_bit()
```

src/sketch.cpp
```diff
@@ -99,17 +146,21 @@ SketchSample Sketch::sample() {
...
     return {buckets[num_buckets - 1].alpha, GOOD};
 
-  for (size_t i = 0; i < cols_per_sample; ++i) {
-    for (size_t j = 0; j < bkt_per_col; ++j) {
-      size_t bucket_id = (i + first_column) * bkt_per_col + j;
-      if (Bucket_Boruvka::is_good(buckets[bucket_id], checksum_seed()))
-        return {buckets[bucket_id].alpha, GOOD};
+
+  for (size_t col = first_column; col < first_column + cols_per_sample; ++col) {
+    int row = effective_size(col)-1;
```
etwest: `int(effective_size(col))`

src/sketch.cpp
```diff
@@ -99,17 +146,21 @@ SketchSample Sketch::sample() {
   size_t idx = sample_idx++;
   size_t first_column = idx * cols_per_sample;
 
-  if (buckets[num_buckets - 1].alpha == 0 && buckets[num_buckets - 1].gamma == 0)
+  if (Bucket_Boruvka::is_empty(buckets[num_buckets - 1]))
     return {0, ZERO};  // the "first" bucket is deterministic so if all zero then no edges to return
 
   if (Bucket_Boruvka::is_good(buckets[num_buckets - 1], checksum_seed()))
```
etwest: Move the `is_good()` check below the for loop because most of the time we don't need to check it. And if we do need to check it the for loop is inexpensive.

src/sketch.cpp
```diff
@@ -124,20 +175,22 @@ ExhaustiveSketchSample Sketch::exhaustive_sample() {
   size_t idx = sample_idx++;
   size_t first_column = idx * cols_per_sample;
 
-  unlikely_if (buckets[num_buckets - 1].alpha == 0 && buckets[num_buckets - 1].gamma == 0)
+  unlikely_if (Bucket_Boruvka::is_empty(buckets[num_buckets -1]))
     return {ret, ZERO}; // the "first" bucket is deterministic so if zero then no edges to return
 
   unlikely_if (Bucket_Boruvka::is_good(buckets[num_buckets - 1], checksum_seed())) {
```
etwest: Again move this below the for loop.

include/sketch.h
```diff
@@ -125,6 +125,18 @@ class Sketch {
...
+   * Gets the 1-index of the deepest non-zero bucket. 
+   * @return  The depth of the non-zero'th bucket + 1. If the bucket is entirely empty, returns 0
+  */
+  uint8_t effective_size(size_t col_idx) const;
+
+  #ifdef EAGER_BUCKET_CHECK
+  vec_t *nonempty_buckets;
+  void update_flags(size_t col_idx, size_t start_row, size_t end_row);
```
etwest: Document me pls

etwest: Also, make all of this private

etwest: `recalculate_flags()`

src/sketch.cpp
```diff
@@ -146,12 +199,46 @@ ExhaustiveSketchSample Sketch::exhaustive_sample() {
...
 void Sketch::merge(const Sketch &other) {
-  for (size_t i = 0; i < num_buckets; ++i) {
-    buckets[i].alpha ^= other.buckets[i].alpha;
-    buckets[i].gamma ^= other.buckets[i].gamma;
+  for (size_t i=0; i < num_columns; ++i) {
+    Bucket *current_col = buckets + (i* bkt_per_col);
+    Bucket *other_col = other.buckets + (i * bkt_per_col);
+    size_t other_effective_size = other.effective_size(i);
+    // size_t other_effective_size = bkt_per_col;
+    #pragma omp simd
```
etwest, line 275: `#pragma`/`#ifdef` should not have indent

src/sketch.cpp
```diff
@@ -146,12 +199,46 @@ ExhaustiveSketchSample Sketch::exhaustive_sample() {
...
+    for (size_t bucket_id=0; bucket_id < other_effective_size; bucket_id++) {
+      current_col[bucket_id].alpha ^= other_col[bucket_id].alpha;
+      current_col[bucket_id].gamma ^= other_col[bucket_id].gamma;
```
etwest: Performance thing to check: Would it be better to perform a bitwise OR of the flags and then after each merge perform an `unlikely_if` to check if the bucket is now empty. If so, clear the flag.

src/sketch.cpp
```diff
@@ -146,12 +199,46 @@ ExhaustiveSketchSample Sketch::exhaustive_sample() {
...
+
+#ifdef EAGER_BUCKET_CHECK
+void Sketch::update_flags(size_t col_idx, size_t start_idx, size_t end_idx) {
+  Bucket *current_col = buckets + (col_idx * bkt_per_col);
+  assert(end_idx >= start_idx);
+  vec_t clear_mask = (~0) >> (8*sizeof(vec_t) - (end_idx - start_idx));
+  clear_mask = ~(clear_mask << start_idx);
+  vec_t good_buck_status = 0;
+  #pragma omp simd
+  for (size_t bucket_id=start_idx; bucket_id < end_idx; bucket_id++) {
```
etwest, line 320: start off with good_buck_status being all 1s and clear it in an unlikely_if - Gil

src/sketch.cpp
```diff
@@ -146,12 +199,46 @@ ExhaustiveSketchSample Sketch::exhaustive_sample() {
   return {ret, GOOD};
 }
 
+
 void Sketch::merge(const Sketch &other) {
```
etwest, line 268: Should we make this function call `range_merge()`

src/sketch.cpp
```diff
@@ -146,12 +199,46 @@ ExhaustiveSketchSample Sketch::exhaustive_sample() {
...
+    // good_buck_status |= (!Bucket_Boruvka::is_empty(current_col[bucket_id])) << bucket_id;
+    likely_if(!Bucket_Boruvka::is_empty(current_col[bucket_id])) set_bit(good_buck_status, bucket_id);
   }
+  nonempty_buckets[col_idx] = (nonempty_buckets[col_idx] & clear_mask) | good_buck_status;
 }
+#endif
+
+
 
 void Sketch::range_merge(const Sketch &other, size_t start_sample, size_t n_samples) {
```
etwest, line 331: Use `effective_size()`?

src/sketch.cpp
```diff
@@ -176,17 +263,67 @@ void Sketch::range_merge(const Sketch &other, size_t start_sample, size_t n_samp
...
   }
+  #ifdef EAGER_BUCKET_CHECK
+  for (size_t col_idx=0; col_idx < num_columns; col_idx++) {
+    update_flags(col_idx, 0, bkt_per_col);
+  }
+  #endif
 }
 
+uint8_t Sketch::effective_size(size_t col_idx) const
+{
```
etwest, line 389: Our style guide is same line {

src/sketch.cpp
```diff
@@ -176,17 +263,67 @@ void Sketch::range_merge(const Sketch &other, size_t start_sample, size_t n_samp
...
+  // first, check for emptyness
+  Bucket *current_row = buckets + (col_idx * bkt_per_col);
+  if (Bucket_Boruvka::is_empty(buckets[num_buckets - 1]))
```
etwest: I would drop this or move it into the `#else`. It's a more expensive call than the `clzll`

src/sketch.cpp
```diff
@@ -176,17 +263,67 @@ void Sketch::range_merge(const Sketch &other, size_t start_sample, size_t n_samp
...
+  {
+    return 0;
+  }
+#ifdef EAGER_BUCKET_CHECK
+  unlikely_if(nonempty_buckets[col_idx] == 0) return 0;
+  return (uint8_t)((sizeof(unsigned long long) * 8) - __builtin_clzll(nonempty_buckets[col_idx]));
```
etwest, line 397: Is there a convenient way to use `ctzll`? It's about 3 times faster than `clzll`

tools/benchmark/graphcc_bench.cpp
```diff
@@ -17,7 +17,11 @@
 
 constexpr uint64_t KB = 1024;
```
etwest, line 18: Pls clean this file. (i.e. get rid of the HashSet etc.)

tools/benchmark/graphcc_bench.cpp
```diff
@@ -534,4 +597,215 @@ static void BM_Parallel_DSU_Root(benchmark::State& state) {
...
+      benchmark::DoNotOptimize(elm);
+    }
+  }
+
+  state.counters["Scan_Latency"] = benchmark::Counter(
+      state.iterations(), benchmark::Counter::kIsRate | benchmark::Counter::kInvert);
+}
+BENCHMARK(BM_Std_Set_Hash_Iterator)->RangeMultiplier(2)->Range(1, 1 << 14);
+
+BENCHMARK_MAIN();
```
etwest, line 638: Add newline

src/sketch.cpp
```diff
@@ -5,6 +5,15 @@
 #include <vector>
 #include <cassert>
 
+
+inline static void set_bit(vec_t &t, int position) {
```
etwest, line 9: Suggest switching the position of first bucket to most significant bit. 
`t |= 1 << (sizeof(vec_t) * 8 - position)`

etwest (changes requested):

include/sketch.h
Gillgamesh: couldn't think of one

